PR #2, opened 2023-05-10
lspserver: document sync framework

inclyc:
Implemented text document sync methods, incrementally sync the textDocument. The implementation is picked from [clangd](https://github.com/llvm/llvm-project/blob/63eb04a3683996db26dbf1534682c5696d93b080/clang-tools-extra/clangd/SourceCode.cpp#L149).

### DIFF
--- a/lib/lspserver/include/lspserver/DraftStore.h
+++ b/lib/lspserver/include/lspserver/DraftStore.h
@@ -1,0 +1,59 @@
+//===--- DraftStore.h - File contents container -----------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "Path.h"
+#include "llvm/ADT/StringMap.h"
+#include "llvm/Support/VirtualFileSystem.h"
+#include <mutex>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace lspserver {
+
+/// A thread-safe container for files opened in a workspace, addressed by
+/// filenames. The contents are owned by the DraftStore.
+/// Each time a draft is updated, it is assigned a version. This can be
+/// specified by the caller or incremented from the previous version.
+class DraftStore {
+public:
+  struct Draft {
+    std::shared_ptr<const std::string> Contents;
+    std::string Version;
+  };
+
+  /// \return Contents of the stored document.
+  /// For untracked files, a std::nullopt is returned.
+  std::optional<Draft> getDraft(PathRef File) const;
+
+  /// \return List of names of the drafts in this store.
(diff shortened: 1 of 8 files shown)


lib/lspserver/include/lspserver/DraftStore.h
```diff
@@ -0,0 +1,59 @@
+//===--- DraftStore.h - File contents container -----------------*- C++ -*-===//
```
junaire, line 1: If we copy & paste this from LLVM monorepo, we should add a link so everyone knows it when they see the code. In addition, it makes it easier to sync with the upstream if anything changes.

lib/lspserver/include/lspserver/SourceCode.h
```diff
@@ -0,0 +1,76 @@
...
+#include "llvm/ADT/StringSet.h"
+#include "llvm/Support/Error.h"
+#include <optional>
+#include <string>
+
+namespace lspserver {
+
+template <class Type> class Key {
+public:
+  static_assert(!std::is_reference<Type>::value,
```
junaire, line 26:
```suggestion
  static_assert(!std::is_reference_v<Type>,
```

lib/nixd/src/Server.cpp
```diff
@@ -23,4 +25,48 @@ void Server::onInitialize(const lspserver::InitializeParams &InitializeParams,
   Reply(std::move(Result));
 }
 
+std::string Server::encodeVersion(std::optional<int64_t> LSPVersion) {
+  return LSPVersion ? llvm::to_string(*LSPVersion) : "";
+}
+
+std::shared_ptr<const std::string>
+Server::getDraft(lspserver::PathRef File) const {
+  auto Draft = DraftMgr.getDraft(File);
```
junaire, line 34:
```suggestion
  std::shared_ptr<const std::string> Draft = DraftMgr.getDraft(File);
```

lib/nixd/src/Server.cpp
```diff
@@ -23,4 +25,48 @@ void Server::onInitialize(const lspserver::InitializeParams &InitializeParams,
...
+
+  const std::string &Contents = Params.textDocument.text;
+
+  addDocument(File, Contents, encodeVersion(Params.textDocument.version));
+}
+
+void Server::onDocumentDidChange(
+    const lspserver::DidChangeTextDocumentParams &Params) {
+  lspserver::PathRef File = Params.textDocument.uri.file();
+  auto Code = getDraft(File);
```
junaire, line 52:
```suggestion
  std::shared_ptr<const std::string> Code = getDraft(File);
```

junaire (approved):

lib/lspserver/include/lspserver/DraftStore.h
inclyc: We can provide a link here: https://github.com/inclyc/nixd/blob/main/lib/lspserver/README.md

However, using git submodule & sync with upstream is not applicable, because clangd has it own logic about how to index & parsing files, coupled with the `DraftStore`, `SourceCode`, etc.

inclyc:
reverted "auto" comments because of CI failure

PR #3, opened 2023-05-11
lang: parse nix file using libexpr on document open

inclyc:


lib/nixd/src/Server.cpp
```diff
@@ -44,6 +56,20 @@ void Server::onDocumentDidOpen(
   const std::string &Contents = Params.textDocument.text;
 
   addDocument(File, Contents, encodeVersion(Params.textDocument.version));
+  lspserver::elog("nix settings: {0}", nix::settings.storeUri.get());
+  auto NixStore = nix::openStore();
+  auto State = std::make_unique<nix::EvalState>(nix::Strings(), NixStore);
+  try {
+    State->parseExprFromFile(File.str());
+  } catch (const nix::ParseError &PE) {
+    lspserver::elog("got nix parse error! {0}", PE.msg());
```
inclyc, line 65: Ideally, we can catch nix exceptions & retrieve backtrace in this exception (provided by nix itself).

lib/nixd/src/Server.cpp
```diff
@@ -44,6 +56,20 @@ void Server::onDocumentDidOpen(
   const std::string &Contents = Params.textDocument.text;
 
   addDocument(File, Contents, encodeVersion(Params.textDocument.version));
+  lspserver::elog("nix settings: {0}", nix::settings.storeUri.get());
+  auto NixStore = nix::openStore();
+  auto State = std::make_unique<nix::EvalState>(nix::Strings(), NixStore);
```
inclyc, line 61: However, this statement is throwing exceptions in `EvalState` constructor?

tools/nixd.cpp
```diff
@@ -9,10 +9,6 @@
 
 int main() {
   std::error_code EC;
-  auto DebugOut = llvm::raw_fd_ostream("/tmp/nixd.out", EC,
```
inclyc, line 12: Currently we have initialized with the VSCode extension. So logging into `stderr` could be transferred into extension outputs. So delete hard-coded log file. (This should be another patch though)

inclyc:
We need to update the build system to include `config.h`. Otherwise record declarations are different from "nix vendored" version.

PR #6, opened 2023-05-12
lsp: binder for method calls & notification

inclyc:


lib/lspserver/include/lspserver/LSPBinder.h
```diff
@@ -9,6 +9,28 @@
 
 namespace lspserver {
 
+template <typename T>
+llvm::Expected<T> parseParam(const llvm::json::Value &Raw,
```
inclyc, line 13: renamed this as global function

lib/lspserver/include/lspserver/LSPBinder.h
```diff
@@ -82,98 +81,12 @@ struct HandlerRegistry {
...
-      auto P = parse<Param>(RawParams, Command, "command");
+      auto P = parseParam<Param>(RawParams, Command, "command");
       if (!P)
         return Reply(P.takeError());
       (This->*Handler)(*P, std::move(Reply));
     };
   }
 };
 
-/// LSPBinder collects a table of functions that handle LSP calls.
```
inclyc, line 93: The functions are replaced by lambdas

lib/lspserver/include/lspserver/LSPServer.h
```diff
@@ -16,12 +23,57 @@ class LSPServer : public MessageHandler {
...
+    llvm::json::Value ID{bindReply(std::move(CB))};
+    Out->call(Method, Params, ID);
   }
 
 protected:
   HandlerRegistry Registry;
+  template <class T>
+  llvm::unique_function<void(const T &)>
+  mkOutNotifiction(llvm::StringRef Method) {
+    return [Method = Method, this](const T &Params) {
```
inclyc, line 58:
```suggestion
    return [Method, this](const T &Params) {
```

?

lib/lspserver/src/LSPServer.cpp
```diff
@@ -1,6 +1,10 @@
 #include "lspserver/LSPServer.h"
 #include "lspserver/Connection.h"
+#include "lspserver/Function.h"
```
inclyc, line 3: nit: split header

PR #17, opened 2023-05-18
nixd: add scheduler support for multi-threading

inclyc:


lib/nixd/include/nixd/Scheduler.h
```diff
@@ -0,0 +1,109 @@
...
+  Task &operator=(const Task &) = delete;
+  Task(Task &&) = default;
+  Task &operator=(Task &&) = default;
+
+  virtual ~Task() = default;
+
+  virtual void run() noexcept = 0;
+};
+
+class EvaluationTask : public Task {
```
inclyc, line 39: This struct is not tested

lib/nixd/src/Scheduler.cpp
```diff
@@ -0,0 +1,40 @@
+#include "nixd/Scheduler.h"
+
+namespace nixd {
+
+void EvaluationTask::run() noexcept {
```
inclyc, line 5: not tested

inclyc:
close because this is buggy, switch to `boost::asio::thread_pool` now.

PR #18, opened 2023-05-19
nixd: add AST library

inclyc:


lib/nixd/test/ast.cpp
```diff
@@ -0,0 +1,47 @@
+#include <exception>
+#include <gtest/gtest.h>
+
+#include "installable-value.hh"
```
inclyc, line 4:
```suggestion
#include "<nix/installable-value.hh>"
```

lib/nixd/include/nixd/AST.h
```diff
@@ -0,0 +1,102 @@
...
+  /// nix 'Env's contains dynamic variable name bindings at evaluation, might be
+  /// used for completion.
+  nix::Env getEnv(nix::Expr *Expr) {
+    assert(Evaluated && "must be called after evaluation!");
+    return EnvMap[Expr];
+  }
+
+  /// Lookup an AST node located at the position.
+  /// Runs in O(n) (inefficient!)
+  nix::Expr *lookupPosition(const nix::EvalState &State, lspserver::Position);
```
inclyc, line 73: definition?

lib/nixd/test/ast.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "nixd/AST.h"
+#include "nixd/Expr.h"
+#include "nixutil.h"
```
inclyc, line 7: same

lib/nixd/test/ast.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "value.hh"
```
inclyc, line 8: same

lib/nixd/include/nixd/AST.h
```diff
@@ -0,0 +1,102 @@
...
+  nix::Expr *Root;
+  std::map<const nix::Expr *, nix::Value> ValueMap;
+  std::map<const nix::Expr *, nix::Env> EnvMap;
+
+public:
+  /// Inject myself into nix cache.
+  void injectAST(nix::EvalState &State, lspserver::PathRef Path);
+
+  /// From fresh-parsed nix::Expr *AST root, e.g. State->parseExprFromString
+  NixAST(ASTContext &Cxt, nix::Expr *Root);
```
junaire, line 55: Ctx?

lib/nixd/include/nixd/AST.h
```diff
@@ -0,0 +1,102 @@
...
+
+  /// Get the evaluation result (fixed point) of the expression.
+  nix::Value getValue(nix::Expr *Expr) {
```
junaire, line 58: What happens if `ValueMap[Expr]` doesn't exist? return an error? or just assert.

lib/nixd/include/nixd/AST.h
```diff
@@ -0,0 +1,102 @@
...
+  /// evaluation. Evaluation errors shall be passed into the first argument.
+  void withEvaluation(
+      Scheduler &Scheduler, std::string Name,
+      nix::ref<nix::InstallableValue> IValue, nix::ref<nix::EvalState> State,
+      llvm::unique_function<void(std::exception *, NixAST *)> Result);
+};
+
+/// Parsing ASTs on each document version, providing ASTs & evaluation analysis
+/// on requests
+class ASTManager {
```
junaire, line 88: Not sure if it's worth it to make it a class. or current implementation just a draft and incomplete?

junaire (commented):
I can't understand everything but just give my 2 cents, will play with the code once I have more bandwidth.

lib/nixd/include/nixd/AST.h
inclyc: NYI

lib/nixd/include/nixd/AST.h
inclyc: nice catch! I think should **assert** here, because `getValue` is expected to be called with ASTNodes managed by the instance, e.g. returned by `NixAST::lookup`

junaire: should we return a reference iiuc?

PR #24, opened 2023-05-26
nixd: fix exception

euphgh:
1. add std::filesystem
2. add const for exception
3. fix hover exception


inclyc:
This has been fixed already.

Issue #32, opened 2023-05-29
nixd/regression: fix arrow direction in regression tests


inclyc:
Some regression tests, like

https://github.com/inclyc/nixd/blob/2ee1ef796dd6dd5f0bc8cc4a3bc804972018c973/tools/nixd/test/basic-diagnostic.md?plain=1#L3

The arrow should be `<--`

JayPankajPatel:
Hey could you assign this task to me?

PR #34, opened 2023-05-30
nixd/regression: fixed arrow direction

JayPankajPatel:
Fixes: #32 

Fixed a mistake where the arrow was not facing the same direction as the other arrows in the rest of the file.

### DIFF
--- a/tools/nixd/test/basic-diagnostic.md
+++ b/tools/nixd/test/basic-diagnostic.md
@@ -1,6 +1,6 @@
 # RUN: nixd --lit-test < %s | FileCheck %s
 
---> initialize(0)
+<-- initialize(0)
 
 ```json
 {
--- a/tools/nixd/test/initialize.md
+++ b/tools/nixd/test/initialize.md
@@ -2,7 +2,7 @@
 
 Check basic handshake with the server, i.e. "initialize"
 
---> initialize(0)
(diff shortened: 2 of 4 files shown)


inclyc:
Thanks for contributing :)

Actually there are exactly same mistakes (i.e. incorrect arrow direction for `initialize(0)`) in our tests (I just employed copy-paste to write these regression tests). Can you fix them all in this PR? e.g.

https://github.com/inclyc/nixd/blob/2ee1ef796dd6dd5f0bc8cc4a3bc804972018c973/tools/nixd/test/initialize.md?plain=1#L5

> facing the same direction as the other arrows in the rest of the file

The directions is not supposed to be identical, instead it denotes the direction LSP server interacting with clients. For example, the messages with `<--` means it was sent by the client to the server (I've not documented this yet).

JayPankajPatel:
Yes I’ll be happy too!
Thanks for the feedback, I’ll be careful not to make assumptions. Yeah it’s clear it is a server interaction to be frank I didn’t read much into but, I’ll be more careful and mindful.

inclyc (approved):

JayPankajPatel:
Fixes: #32 

Added fix to all affected files.

JayPankajPatel:
Thanks for being my first contribution to an open-source project!

Issue #35, opened 2023-05-30
vendor: package this repository in `flake.nix`


inclyc:
As mentioned in

https://github.com/inclyc/nixd/blob/72d2bdbb58c02f4d9a149737bd7febdb7548a979/README.md?plain=1#L63

We should make a derivation in our `flake.nix`, and let our CI perform something like: `nix build .#`.

PR #41, opened 2023-05-31
vendor: add a default.nix expression

adisbladis:
Fixes: #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,13 +60,22 @@ And diagnostic:
 
 ### Build the project
 
-This project is working in progress, and does not vendored dedicated derivation yet.
-However, if you'd like to help us improve this work, please enter the devShell, and build this project via meson.
+#### nix-build
+``` sh
+nix-build --expr 'with import <nixpkgs> { }; callPackage ./. { }'
+```
 
-Here are a short snippet used in our CI, should works fine and reproducible in your environment.
+#### Nix Flakes
+``` sh
+nix build -L #.
+```
+
+#### Development
+
+Here are a short snippet used in our CI, should works fine and reproducible in your development environment.
 
 ```
-meson setup build/ --buildtype=${{ matrix.buildtype }} -Db_sanitize=${{ matrix.sanitizer }} -Db_ndebug=${{ matrix.ndebug }}
+meson setup build/
 meson compile -C build
 meson test -C build
 ```
--- a/flake.nix
+++ b/flake.nix
@@ -5,7 +5,7 @@
     flake-parts.url = "github:hercules-ci/flake-parts";
   };
 
-  outputs = { flake-parts, ... }@inputs: flake-parts.lib.mkFlake { inherit inputs; } {
+  outputs = { nixpkgs, flake-parts, ... }@inputs: flake-parts.lib.mkFlake { inherit inputs; } {
     perSystem = { config, self', inputs', pkgs, system, ... }:
       let
         nixd = pkgs.callPackage ./default.nix { };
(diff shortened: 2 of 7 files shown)


lib/nixd/include/nixd/AST.h
```diff
@@ -8,6 +8,7 @@
 #include <llvm/ADT/FunctionExtras.h>
 #include <llvm/ADT/StringRef.h>
 
+#include <nix/config.h>
```
inclyc, line 11: Our headers is **sorted** in lexicographic order which basically assumes all headers are self-contained. However, `<nix/config.h>` must be included before any "real" nix headers because it contains many critical macro definition, and place it after a nix header file may lead to incorrect struct layout. (If I'm wrong please comment)

```suggestion
#include <nix/config.h>

```

Let's leave a blank line here, to make sure it is not sorted with the rest nix headers?

---

Also, may I ask why should we do this in C++ source files? (As comment above, `<nix/config.h>` is semantically differently from the rest.

lib/nixd/include/nixd/Diagnostic.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include "lspserver/Protocol.h"
 
+#include <nix/config.h>
```
inclyc, line 4: same

lib/nixd/include/nixd/Expr.h
```diff
@@ -1,6 +1,7 @@
 /// 'nix::Expr' wrapper that suitable for language server
 #pragma once
 
+#include <nix/config.h>
```
inclyc, line 4: same

flake.nix
```diff
@@ -7,43 +7,21 @@
...
+        };
+
+        devShells.default = nixd.overrideAttrs (old: {
+          nativeBuildInputs = old.nativeBuildInputs ++ [ pkgs.clang-tools pkgs.gdb ];
           shellHook = ''
-            export NIX_DEBUG_INFO_DIRS=${nixUnstable.debug}/lib/debug
-            export NIX_SRC=${nixUnstable.src}
-            export NIX_CONFIG_H=${nixUnstable.dev}/include/nix/config.h
-            export CXXFLAGS="-include $NIX_CONFIG_H"
+            export NIX_DEBUG_INFO_DIRS=${pkgs.nixUnstable.debug}/lib/debug
```
inclyc, line 22: I'd like to keep `NIX_SRC` here for my debugging experience :)

README.md
```diff
@@ -60,13 +60,22 @@ And diagnostic:
 
 ### Build the project
 
-This project is working in progress, and does not vendored dedicated derivation yet.
-However, if you'd like to help us improve this work, please enter the devShell, and build this project via meson.
+#### nix-build
```
inclyc, line 63: Also add a CI file for this stuff?

lib/nixd/include/nixd/AST.h
adisbladis: I wanted to do it in the cpp source files as that would be resolved normally by the compiler and we don't have to mess around with compiler include ptahs.

In any case I think that this should probably be considered a bug in the Nix headers.

We could just fix the issue there and then drop the `-include` at a later point.

inclyc: > In any case I think that this should probably be considered a bug in the Nix headers.

Frankly Eelco did the same hack here:

https://github.com/edolstra/dwarffs/blob/1f850df9c932acb95da2f31b576a8f6c7c188376/flake.nix#L24

So this might be a "feature" for nix itself?

adisbladis:
@inclyc This should be good to review.

inclyc (approved):
Thanks!

573:
Thanks @adisbladis and @inclyc

PR #45, opened 2023-05-31
nixd/eval: put this stuff in main thread

inclyc:
... This will dramatically affect our speed on responding requests.

However, we have to do this, because nix evaluation is NOT thread-safe.


inclyc:
I won't change the structure of current codebase, because this is just a workaround for threaded evaluation.

How can I made evaluation in a detached thread, with out any race condition introduced by GC?

PR #50, opened 2023-06-01
nixd/fork: forking concurrency

inclyc:
I'd like to eval workspace files in a separated process. When we update the workspace, we call `fork()` to lazily duplicate nixd itself. And, the server turns itself into `Worker` mode, which means:

1. Enable language feature by calling nix::init*, and also start the GC.
2. Do not response workspace update request, just fixing the workspace, with evaluation runs at once.
3. Handle LSP features, like `Hover`
4. The Worker itself runs in single-thread mode.

The main process it self redirects LSP requests to versioned workers, and just choose the best result, and thus should be cancellable.

The workers are managed in a deque, which means we can kill outdated processes (and thus free the memory).

regression status:

- [x] diagnostic
- [x] completion
- [x] hover


lib/nixd/include/nixd/Server.h
```diff
@@ -13,8 +13,29 @@
...
+
+bool fromJSON(const llvm::json::Value &Params, CompletionItem &R,
+              llvm::json::Path P);
+
+bool fromJSON(const llvm::json::Value &Params, CompletionList &R,
+              llvm::json::Path P);
+
+} // namespace lspserver
+
+namespace nixd { // namespace nixd
```
inclyc, line 38: ?

lib/lspserver/src/Logger.cpp
```diff
@@ -51,8 +51,8 @@ void StreamLogger::log(Logger::Level Level, const char *Fmt,
     return;
   llvm::sys::TimePoint<> Timestamp = std::chrono::system_clock::now();
   std::lock_guard<std::mutex> Guard(StreamMutex);
-  Logs << llvm::formatv("{0}[{1:%H:%M:%S.%L}] {2}\n", indicator(Level),
-                        Timestamp, Message);
+  Logs << llvm::formatv("{0}[{1:%H:%M:%S.%L}] {2}: {3}\n", indicator(Level),
+                        Timestamp, getpid(), Message);
```
inclyc, line 55: `#include` `unistd.h`

inclyc:
Now all features are implemented in forking architecture, however looks like something is not working in github ci, but I cannot reproduce locally.

lib/nixd/include/nixd/Server.h
```diff
@@ -49,11 +75,80 @@ struct CompletionHelper {
...
+    std::unique_ptr<nix::Pipe> ToPipe;
+    std::unique_ptr<nix::Pipe> FromPipe;
+    std::unique_ptr<lspserver::OutboundPort> OutPort;
+    std::unique_ptr<llvm::raw_ostream> OwnedStream;
+
+    nix::Pid Pid;
+    WorkspaceVersionTy WorkspaceVersion;
+    std::thread InputDispatcher;
+
+    Proc(decltype(ToPipe) ToPipe, decltype(FromPipe) FromPipe,
```
inclyc, line 132: How about rewrite this stuff in InitListExpr ?

junaire (approved):

Issue #52, opened 2023-06-02
Semantic Tokens


junaire:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens

inclyc:
This feature is non-trivial before we have something range-based AST library. Closing as "not planned" here basically means it is too hard to implement. This feature is definitely "planned" to impl in the future though.

PR #57, opened 2023-06-03
nixd/semantic-tokens: init

junaire:
Fixes: https://github.com/nix-community/nixd/issues/52


inclyc:
Looks like this is out-dated for a long time.

PR #62, opened 2023-06-03
nixd/expr: lookup position support (low-level)

### LINKED ISSUE #61: nixd: enhance AST library, for location lookup & search up
- [x] low-level `nix::Expr *` support #62 
- [ ] `nixd::EvalAST` #65

inclyc:
This partially fixes

    #61 (nixd: enhance AST library, for location lookup & search up)

Add the low-level operations on nix::Expr*, that could perform location lookup.


inclyc:
GCC does not accept my impl now, but clang said it's OK, let's fix it then.

PR #65, opened 2023-06-04
nixd/AST/def: high-level AST library support for goto * (definition, declaration, ...)

inclyc:
Fixes: #61

### DIFF
--- a/lib/nixd/src/AST.cpp
+++ b/lib/nixd/src/AST.cpp
@@ -31,19 +31,27 @@ void EvalAST::injectAST(nix::EvalState &State, lspserver::PathRef Path) {
 }
 
 nix::Value EvalAST::searchUpValue(const nix::Expr *Expr) const {
-  for (const auto *Parent = parent(Expr); Parent != Expr; Expr = parent(Expr)) {
+  for (;;) {
     if (ValueMap.contains(Expr))
       return ValueMap.at(Expr);
+    if (parent(Expr) == Expr)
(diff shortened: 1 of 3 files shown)


lib/nixd/src/AST.cpp
```diff
@@ -29,6 +30,22 @@ void EvalAST::injectAST(nix::EvalState &State, lspserver::PathRef Path) {
   State.cacheFile(Path.str(), Path.str(), Root, DummyValue);
 }
 
+nix::Value EvalAST::searchUpValue(const nix::Expr *Expr) const {
+  for (const auto *Parent = parent(Expr); Parent != Expr; Expr = parent(Expr)) {
```
inclyc, line 34: Looks like we cannot get env in root node ?

lib/nixd/src/AST.cpp
```diff
@@ -29,6 +30,22 @@ void EvalAST::injectAST(nix::EvalState &State, lspserver::PathRef Path) {
...
+    if (ValueMap.contains(Expr))
+      return ValueMap.at(Expr);
+  }
+  throw std::out_of_range("No such value associated to ancestors");
+}
+
+const nix::Env *EvalAST::searchUpEnv(const nix::Expr *Expr) const {
+  for (const auto *Parent = parent(Expr); Parent != Expr; Expr = parent(Expr)) {
```
inclyc, line 42: same

PR #78, opened 2023-06-05
nixd/format: support formatting (external command `nixpkgs-fmt`)

junaire:


inclyc (changes requested):
Currently there is no error-handling, so I'd like mark this draft again.

inclyc (approved):

Issue #63, opened 2023-06-04
nixd: allow workspace configuration in root dir


inclyc:
nvim users expected "workspace configuration" in their "rootDir", not from LSP client.

inclyc:
How about `.nixd.json` in root dir ?

PR #89, opened 2023-06-08
nixd/config: accept config file from `.nixd.json`

inclyc:
Fixes: #63

### DIFF
--- a/lib/nixd/include/nixd/Server.h
+++ b/lib/nixd/include/nixd/Server.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lspserver/Protocol.h"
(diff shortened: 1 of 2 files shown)


lib/nixd/include/nixd/Server.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lspserver/Protocol.h"
```
inclyc, line 3: extra header

lib/nixd/src/ServerController.cpp
```diff
@@ -1,3 +1,4 @@
+#include "error.hh"
```
inclyc, line 1: extra header

PR #98, opened 2023-06-10
vendor: update nixd package according to reviews from nixpkgs

inclyc:


inclyc:
@adisbladis Would you like to become the code owner of this file? (or something vendor related)

adisbladis:
> @adisbladis Would you like to become the code owner of this file? (or something vendor related)

No. I'm not sure how much involvement I'm going to have here.

Issue #100, opened 2023-06-10
docs/editor: add editor setup guide


domenkozar:
Most users would like to know just one thing: how to setup their editor to use nixd.

It would be good to document that in the README.md, in my case VSCode.

inclyc:
Thanks, use the plugin Nix IDE and set the server path to `nixd` shall work out of box.

> It would be good to document that in the README.md, in my case VSCode.

I would like to do that today.

inclyc:
https://discourse.nixos.org/t/nixd-nix-language-server/28910/4?u=inclyc

PR #54, opened 2023-06-03
improve devShells

Ruixi-rebirth:
Format
- *.nix formatted by nixpkgs-fmt
- *.cpp formatted by clang-format
- *.md, *.yml formatted by prettier

Improve devShells
- add overlay support
- development environment

Provides a method for configuring neovim lsp
- see [upstream docs](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.txt#nixd)

image

PR #102, opened 2023-06-10
docs/editor-setup: add the doc

inclyc:
Fixes: #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ And diagnostic:
 
 ## Resources
 
+- [Editor Setup](docs/editor-setup.md)
 - [User Guide](docs/user-guide.md)
 - [Developers' Manual](docs/dev.md) (internal design, contributing):
 - Project matrix room: https://matrix.to/#/#nixd:matrix.org
--- a/docs/editor-setup.md
+++ b/docs/editor-setup.md
@@ -1,3 +1,7 @@
+## We ❤️ Contributions!
+
+Please update this doc (tweaks, tricks, or how to setup for another editor)!
+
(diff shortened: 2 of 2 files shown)


ncfavier:
Could you please add instructions for Vim/Neovim?

inclyc:
> Could you please add instructions for Vim/Neovim?

I'm not using vim/neovim :)

I do think adding such document will be very nice! However, thanks to my friend @Ruixi-rebirth who added official nvim-lspconfig upstream support for nixd: https://github.com/neovim/nvim-lspconfig/pull/2644.

If you have such things working on your editor, I will be very glad to accept & merge the PR, see https://github.com/nix-community/nixd/blob/db818511c22da2a898f8075f6e1d74e15d960fb0/docs/editor-setup.md?plain=1#L1-L3

inclyc:
You might want to see: https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.txt#nixd

ncfavier:
That seems perfect, thanks. I think the doc should mention that for Neovim, and more generally explain that `nixd` is an LSP server, so should work out of the box with any LSP client if you just set `nixd` as the server command.

inclyc:
> I think the doc should mention that for Neovim.

Oh thanks, I already prepared a PR which cherry-picks the documentation for nvim from previously closed #54  (so I explicitly set the author to @Ruixi-rebirth ). 

> and more generally explain that nixd is an LSP server, so should work out of the box with any LSP client if you just set nixd as the server command.

❤️

Issue #112, opened 2023-06-12
nixd/definition: support "goto definition" on packages


inclyc:
If a variable is a "derivation", we can bring our user to the position that provides the package, the attribute path is under

".meta.position"

PR #113, opened 2023-06-12
nixd: goto definition for derivations

inclyc:
Fixes: #112

This is a cool feature that brings you to the position where a package defined.

goto-def-pkg-2


lib/nixd/src/ServerWorker.cpp
```diff
@@ -9,6 +9,7 @@
 #include "lspserver/Logger.h"
 #include "lspserver/Protocol.h"
 #include "lspserver/SourceCode.h"
+#include "llvm/Support/Compiler.h"
```
inclyc, line 12: ?

inclyc:
TODO: add tests.

Issue #114, opened 2023-06-12
nixd/completion/options: send example to our user, this might be useful


inclyc:
![options-example](https://github.com/nix-community/nixd/assets/36667224/43e00a8e-c2e6-4598-b188-f5e95d708256)

PR #115, opened 2023-06-12
nixd/completion/options: send example field to our user

inclyc:
Fixes: #114


lib/nixd/include/nixd/nix/Option.h
```diff
@@ -2,13 +2,29 @@
...
 namespace nix::nixd {
 struct OptionInfo {
   std::optional<std::string> Type;
   std::optional<std::string> Description;
+  std::optional<std::string> Example;
+
+  std::string mdDoc() {
+    return llvm::formatv(R"(
+## Example
+
```
inclyc, line 19: how about delete this blank line?

Issue #116, opened 2023-06-12
nixd/definition: goto definition for any evaluated non-pos-discarded values 


inclyc:
Inspired by nil users, thanks. https://github.com/oxalica/nil/issues/9

![lambda-location](https://github.com/nix-community/nixd/assets/36667224/5792da0b-8152-4e51-9b0e-0387b045eeb5)

PR #117, opened 2023-06-12
nixd/definition: goto definition for any evaluated non-pos-discarded values

inclyc:
Fixes: #116
Inspired-by: https://github.com/oxalica/nil/issues/9#issuecomment-1311972565

Actually the current impl is just do the comment. However we do not need to start repl or whatever.

### DIFF
--- a/lib/nixd/src/ServerWorker.cpp
+++ b/lib/nixd/src/ServerWorker.cpp
@@ -14,6 +14,7 @@
 #include <nix/error.hh>
 #include <nix/eval.hh>
 #include <nix/globals.hh>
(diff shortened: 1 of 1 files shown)


lib/nixd/src/ServerWorker.cpp
```diff
@@ -9,12 +9,14 @@
 #include "lspserver/Logger.h"
 #include "lspserver/Protocol.h"
 #include "lspserver/SourceCode.h"
+#include "nixexpr.hh"
```
inclyc: .

PR #125, opened 2023-06-13
nixd/hover: sync hover-eval-error test to ensure workers are finished

inclyc:


inclyc:
We still randomly failing tests because there is no sync guarantee between workers. There must be a more elegant way to do this, for lit testing.

inclyc:
> We still randomly encounter failing tests because there is no sync guarantee between workers. There must be a more elegant way to do this, for lit testing.

#126

PR #127, opened 2023-06-14
vendor: use git revision if possible

inclyc:


default.nix
```diff
@@ -22,7 +22,7 @@ let
 in
 stdenv.mkDerivation {
   pname = "nixd";
-  version = "0.0.1";
+  version = "nightly";
```
inclyc, line 25: Can we get `git describe` in the package script? @Ruixi-rebirth

Ruixi-rebirth: LGTM

Ruixi-rebirth (approved):

PR #130, opened 2023-06-14
nixd/AST: lookup values if it is statically binds to some value

inclyc:


lib/nixd/src/AST.cpp
```diff
@@ -32,6 +32,19 @@ void EvalAST::injectAST(nix::EvalState &State, lspserver::PathRef Path) {
                   DummyValue);
 }
 
+nix::Value EvalAST::getValue(const nix::Expr *Expr) const {
+  if (const auto *EV = dynamic_cast<const nix::ExprVar *>(Expr)) {
+    if (!EV->fromWith) {
+      auto EnvExpr = searchEnvExpr(EV, ParentMap);
+      if (auto EL = dynamic_cast<const nix::ExprLambda *>(EnvExpr)) {
```
inclyc, line 39: Currently it only supports lambdas, however we might extend this function to do more general value-association.

PR #131, opened 2023-06-15
nixd/parser: hack upstream parser to support ranges.

inclyc:


lib/nixd/src/nix/Parser.y
```diff
@@ -0,0 +1,691 @@
+// \$\$ = (.*);
+// { $$$$ = $1;  data->locations[$$$$] = CUR_POS; }
+// d|ata->symbols.create\(([^ )]*)\)
+// [&] () { auto Sym = [&] () { auto Sym = data->symbols.create($1); data->symbolLocations.insert({Sym, CUR_POS}); return Sym; }(); data->symbolLocations.insert({Sym, CUR_POS}); return }()
```
inclyc, line 4: Might be something wrong with this change. Because symbols are "uniquely allocated".

PR #105, opened 2023-06-11
docs/editor: about the nvim setup

Ruixi-rebirth:

PR #132, opened 2023-06-15
editors/nvim-lsp: fix don't display diagnostic information

Ruixi-rebirth:
When the cursor is moved to the problematic line the diagnostic information will be displayed as a floating window

image

### DIFF
--- a/editors/nvim-lsp.nix
+++ b/editors/nvim-lsp.nix
@@ -32,43 +32,43 @@ let
     -- Basic options --
     -------------------
     local options = {
-      clipboard = "unnamedplus",
-      mouse = "a",
-      undofile = true,
-      ignorecase = true,
-      showmode = false,
-      showtabline = 2,
-      smartindent = true,
-      autoindent = true,
-      swapfile = false,
-      hidden = true, --default on
-      expandtab = true,
-      cmdheight = 1,
-      shiftwidth = 2, --insert 2 spaces for each indentation
-      tabstop = 2, --insert 2 spaces for a tab
-      cursorline = true, --Highlight the line where the cursor is located
-      cursorcolumn = false,
-      number = true,
-      numberwidth = 4,
-      relativenumber = true,
-      scrolloff = 8,
-      updatetime = 50, -- faster completion (4000ms default)
+    	clipboard = "unnamedplus",
+    	mouse = "a",
+    	undofile = true,
+    	ignorecase = true,
+    	showmode = false,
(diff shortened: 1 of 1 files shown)


editors/nvim-lsp.nix
```diff
@@ -32,43 +32,43 @@ let
...
-      shiftwidth = 2, --insert 2 spaces for each indentation
-      tabstop = 2, --insert 2 spaces for a tab
-      cursorline = true, --Highlight the line where the cursor is located
-      cursorcolumn = false,
-      number = true,
-      numberwidth = 4,
-      relativenumber = true,
-      scrolloff = 8,
-      updatetime = 50, -- faster completion (4000ms default)
+    	clipboard = "unnamedplus",
```
inclyc, line 35: Formatting changes necessary?

Ruixi-rebirth: There was a problem formatting the lua-related code in this pr #105 , so I reformatted it

573:
Did not find yet in this PR but it might help other users  of the prepared `nvim-lsp`  too to add
`vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, {virtual_text = true})` to the lua configuration.

This way they immediately have feedback what's potentially wrong in their nix file when eval refuses to work for some reason.

573:
When someone is don't want that in their config they can still factor it out but for the sake of a demo I find it reasonable to keep the feedback.

editors/nvim-lsp.nix
inclyc: > There was a problem formatting the lua-related code in this pr https://github.com/nix-community/nixd/pull/105 , so I reformatted it

Ideally you should just push the main branch without review, to reformat your own code. Instead of submit PRs & request review. That might be some noise to the reviewer.

Ruixi-rebirth:
Showing all the diagnostic information directly is a bit annoying and can even make the whole code look confusing. But in this pr, there is always an error icon displayed next to the problematic line, so that you can simply move the cursor to the line where it is and the diagnostic information will be displayed, so that the code will look much cleaner

573:
Hover doesn't work here so I'll have to dig deeper.

inclyc (approved):
LGTM, however as I commented in https://github.com/nix-community/nixd/pull/132#discussion_r1230620884 , please submit formatting changes directly, without review. And formatting changes this from your commit.

Issue #138, opened 2023-06-16
nixd/documentLink: support documentLink on `ExprPath`s


inclyc:
After #136 #131, we have extended the parser that having ranges for all AST nodes (mostly, except ad-hoc created ones). So it will be nice to having documentLink on `ExprPath` s.

PR #140, opened 2023-06-16
nixd/documentLink: init

inclyc:
This is currently not testable, because nix will try to lstat the path.

Fixes: #138


inclyc:
Currently implementation is slow, and might be move to a dedicated parser worker.

PR #156, opened 2023-06-19
nixd: static name lookups for completion

inclyc:


lib/nixd/src/Expr.cpp
```diff
@@ -1,5 +1,7 @@
 #include "nixd/Expr.h"
 
+#include <algorithm>
+#include <iterator>
```
inclyc, line 4: sort

lib/nixd/src/ServerEval.cpp
```diff
@@ -11,6 +11,8 @@
 #include "lspserver/Protocol.h"
 #include "lspserver/SourceCode.h"
 
+#include <algorithm>
```
inclyc, line 14: sort

PR #158, opened 2023-06-20
nixd/eval: automatically force value when user requests

inclyc:


lib/nixd/src/AST.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include "lspserver/Logger.h"
 #include "lspserver/Protocol.h"
+#include "nixd/nix/EvalState.h"
```
inclyc, line 9: nit

PR #161, opened 2023-06-20
docs/README: update "write a package"

inclyc:


Ruixi-rebirth (approved):

573 (commented):
The `write-package` gif still assumes `llvmPackages_15`. Is there a reason to not just use `llvmPackages` ?
Rest looks good for me.

inclyc:
> The `write-package` gif still assumes `llvmPackages_15`. Is there a reason to not just use `llvmPackages` ? Rest looks good for me.

llvmPackages might be too old (is it 11?). I know some people may not have this package set in their channels, but actually nixpkgs has introduced 16 for a while, 15 is a relatively old version. Maybe just `nix-channel --update`?

PR #162, opened 2023-06-20
docs/examples: examples about how to use nixd

inclyc:


inclyc (commented):
Looks like we are missing hm options, should we add the example here?

Home-manager might be installed via flakes or nix-channels, really don't know if it is OK to add a flake here

PR #168, opened 2023-06-22
editors/vscodium: add test env for vscod{e,ium} + Nix IDE

Ruixi-rebirth:


flake.nix
```diff
@@ -47,6 +47,13 @@
             (import ./editors/nvim-lsp.nix { inherit pkgs; })
           ];
         };
+        devShells.vscodium = pkgs.mkShell {
+          nativeBuildInputs = [
+            nixd
+            pkgs.nixpkgs-fmt
```
inclyc, line 53: Maybe we can share the same dependency expression with `devShells.nvim`, and if the dependency is changed in the future, it can be updated it consistently

editors/vscodium.nix
```diff
@@ -0,0 +1,27 @@
...
+  ${coreutils}/bin/mkdir -p "$dir/User"
+  cat >"$dir/User/settings.json" <<EOF
+  {
+  "security.workspace.trust.enabled": false,
+  "nix.enableLanguageServer": true,
+  "nix.serverPath": "nixd",
+  }
+  EOF
+  if [[ "$XDG_SESSION_TYPE"=="wayland" ]];then
+      NIXOS_OZONE_WL=1 ${codium}/bin/codium --disable-gpu --user-data-dir "$dir" "$@"
```
inclyc, line 23: Is this works for all wayland users? I'm concerning about some people on wayland may not launch it correctly, with wayland ozone. Maybe just use `${codium}/bin/codium --user-data-dir "$dir" "$@"`, (i.e. use xwayland) will be better.

inclyc (commented):
And, update the documentation saying that we support this?

editors/vscodium.nix
```diff
@@ -0,0 +1,27 @@
...
+  codium = vscode-with-extensions.override {
+    vscode = vscodium;
+    vscodeExtensions = with vscode-extensions; [
+      jnoortheen.nix-ide
+    ];
+  };
+in
+writeShellScriptBin "codium-test" ''
+  set -e
+  dir="''${XDG_CACHE_HOME:-~/.cache}/nixd-codium"
```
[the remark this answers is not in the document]
Ruixi-rebirth: No, you need to delete it yourself manually

editors/vscodium.nix
Ruixi-rebirth: I think what you say is reasonable, it may not work for all wayland users, if there is a problem, just leave the launch parameter to the user to decide

Ruixi-rebirth:
> And, update the documentation saying that we support this?

Please see #170

PR #169, opened 2023-06-22
docs/README: use some nix-style for our title

inclyc:


inclyc:
Hmm, github reject to render our custom style :(

Issue #173, opened 2023-06-24
nixd/hover: hover documentation for nixpkgs functions


inclyc:
It might be useful if we can extract comments from nixpkgs, this need a slightly modified lexer.

https://github.com/nix-community/nixd/blob/3dbe1eb6bde1949b510e19a2d1863a2f4d2329a6/lib/nixd/src/nix/Lexer.l#L310

The comments are discarded directly in our lexer now, we need to preserve this information and a dedicated structure for comments.

---

Another problem is, the location information is not correct for a Currying lambda. e.g. `stdenv.mkDerivation` takes us to the location where the origin lambda defined, not "alias"-d location.

inclyc:
- RFC https://github.com/NixOS/rfcs/pull/145 specified the relationship between doc-comments and expressions.
- We have upstream PR that extract comments and can be used in `:doc` https://github.com/NixOS/nix/pull/9054

PR #176, opened 2023-06-24
nixd/eval/definition: support attr goto definition

inclyc:
goto-attrfield

Bring you where are the attr fields defined, instead of values.

To address the second point mentioned in https://github.com/nix-community/nixd/issues/173


inclyc:
Close as outdated.

PR #183, opened 2023-06-28
nixd-ast-dump: init

inclyc:


inclyc:
darwin, again.

default.nix
```diff
@@ -12,6 +12,7 @@
 , nixVersions
 , nixpkgs-fmt
 , pkg-config
+, valgrind
```
inclyc, line 15: unused

PR #186, opened 2023-06-29
nixd/Parser: make parsing memory-safe for errors, using RAII

inclyc:
Free memory using a series of context object, record all allocated references, and free them automatically.

Previous implementation do require there are no syntax errors. But this one is not leaking for any errors.


nixd/include/nixd/Parser/Require.h
```diff
@@ -1,48 +1,65 @@
 #pragma once
 
+#include "nixd/Expr/Expr.h"
+#include "types.hh"
```
inclyc, line 4: .

nixd/include/nixd/Parser/Prologue.h
```diff
@@ -6,6 +6,7 @@
 #include "Lexer.tab.h"
 
 #include "Provides.h"
+#include "nixd/Parser/Require.h"
```
inclyc, line 9:
```suggestion
#include "Require.h"
```

PR #205, opened 2023-07-03
nixd/completion: move static completion to controller

inclyc:
Fixes: https://github.com/nix-community/nixd/issues/193


nixd/include/nixd/Server/Server.h
```diff
@@ -2,6 +2,7 @@
 
 #include "EvalDraftStore.h"
 
+#include "nixd/Parser/Require.h"
```
inclyc, line 5: .

PR #218, opened 2023-07-09
nixd/Server: join threads in ~Server before ~Proc

inclyc:
This should fix a potential data race on https://github.com/nix-community/nixd/actions/runs/5499350517


inclyc:
WaitWorker is implicitly `true` in lit tests. This might not be the real problem

Issue #215, opened 2023-07-09
Cannot build on macOS due to failing tests


starcraft66:
I can't build nixd on my macbook pro (apple silicon)

I get the following errors during the test phase:

```
[----------] 6 tests from Expr                                                                                                       
[ RUN      ] Expr.Visitor1                                                                                                           
unknown file: Failure                                                                                                                
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.               
[  FAILED  ] Expr.Visitor1 (0 ms)                                                                                                    
[ RUN      ] Expr.Visitor2                                                                                                           
unknown file: Failure                                                                                                                
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.               
[  FAILED  ] Expr.Visitor2 (0 ms)                                                                                                    
[ RUN      ] Expr.CallbackRewrite                                                                                                    
unknown file: Failure                                                                                                                
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.               
[  FAILED  ] Expr.CallbackRewrite (0 ms)                                                                                             
[ RUN      ] Expr.ParentMapLetSearch                                                                                                 
unknown file: Failure                                                                                                                
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.               
[  FAILED  ] Expr.ParentMapLetSearch (0 ms)                                                                                          
[ RUN      ] Expr.ParentMapRecSearch
unknown file: Failure
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.
[  FAILED  ] Expr.ParentMapRecSearch (0 ms)
[ RUN      ] Expr.ParentMapLetNested
unknown file: Failure
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.
[  FAILED  ] Expr.ParentMapLetNested (0 ms)
[----------] 6 tests from Expr (1 ms total)

[----------] 2 tests from Parser
[ RUN      ] Parser.Basic
unknown file: Failure
C++ exception with description "error: cannot set alternative stack: Operation not permitted" thrown in the test body.
[  FAILED  ] Parser.Basic (0 ms)
[ RUN      ] Parser.parse1
[       OK ] Parser.parse1 (0 ms)
```

inclyc:
The ci for darwin builds is passing: https://github.com/nix-community/nixd/actions/runs/5478858846/jobs/9979908723

[...]

PR #226, opened 2023-07-14
build derivation: move lit to buildInputs

evanrichter:
lit is necessary for buildPhase, not just checkPhase

found while working around #215

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,12 +31,12 @@ stdenv.mkDerivation {
   ];
 
   nativeCheckInputs = [
-    lit
     nixpkgs-fmt
   ];
 
   buildInputs = [
     libbacktrace
+    lit
     nix
     gtest
     boost182


inclyc:
Alternatively, should we modify the `checkPhase` specifically for macOS?

evanrichter:
I think we should find the root cause of the test failure before potentially hiding something that should be addressed

Stack switching _could_ be something in nixd that is exercised by the tests, but not in my initial LSP usage. In that case, I would hope for a more robust fix. (For now, my immediate fix in the bug report is fine for me, but maybe not for others)

If just the test runner or scripts are the problem, then we can feature gate tests from macos in the short term, but it's probably not a good long term solution.

inclyc:
As discussed in https://github.com/nix-community/nixd/issues/215, this might be caused by the sandboxes in macOS Ventura. How about just disable these tests (e.g. override checkPhase if stdenv.isDarwin)?

PR #230, opened 2023-07-25
nixd/Server: split worker in a dedicated class

inclyc:
This should fix some async-signal unsafe calls (causing the forked process hangs).


nixd/include/nixd/Server/IPC.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "lspserver/Protocol.h"
```
inclyc, line 3: sort

nixd/lib/Server/Controller.cpp
```diff
@@ -1,12 +1,15 @@
+#include "lspserver/DraftStore.h"
```
inclyc, line 1: sort

nixd/lib/Server/Controller.cpp
```diff
@@ -16,7 +19,9 @@
 #include "lspserver/Protocol.h"
 #include "lspserver/SourceCode.h"
 #include "lspserver/URI.h"
+#include "util.hh"
```
inclyc, line 22: sort

nixd/lib/Server/Controller.cpp
```diff
@@ -16,7 +19,9 @@
...
 
+#include <cstddef>
```
inclyc, line 24: sort

Issue #118, opened 2023-06-12
nixd/eval: progress reporting


algorithmiker:
There is a way to do this via the LSP spec:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress

Other language servers like `rust-analyzer` or `dartls` do this. It's very useful as an "I'm working" signal.

inclyc:
Looks good

PR #237, opened 2023-07-30
nixd/Server/Controller: skip workers if they are busy

inclyc:


nixd/include/nixd/Server/Controller.h
```diff
@@ -338,13 +359,19 @@ auto Controller::askWorkers(
                           Result.takeError());
         }
         Sema->release();
+        {
```
inclyc, line 362: will this leak some stuff here if the worker is being killed?

inclyc:
This is not a good idea, if we want to implement "progress reporting"(#118) as well.

Issue #110, opened 2023-06-11
nixd/completion: send incomplete completion list if it is too large


inclyc:

inclyc:
We need a completion session abstraction to do this, might be somehow difficult for our codebase now.

inclyc:
outdated.

PR #242, opened 2023-08-13
nixd/Sema: filter item prefix for completion builder

inclyc:
Fixes: #110


nixd/lib/Sema/CompletionBuilder.cpp
```diff
@@ -122,14 +121,10 @@ void CompletionBuilder::addEnv(const EvalAST &AST, nix::EvalState &State,
 void CompletionBuilder::addStaticEnv(const nix::SymbolTable &STable,
                                      const nix::StaticEnv &SEnv) {
   for (auto [Symbol, Displ] : SEnv.vars) {
-    std::string Name = STable[Symbol];
```
inclyc, line 125: breaking

nixd/lib/Server/EvalWorker.cpp
```diff
@@ -7,6 +7,7 @@
 #include "nixd/Support/String.h"
 
 #include "lspserver/LSPServer.h"
+#include "nixexpr.hh"
```
inclyc, line 10: .

Issue #243, opened 2023-08-15
linters integration: provide diagnostics from linters


inclyc:
https://discourse.nixos.org/t/list-of-nix-linters/19279

PR #245, opened 2023-08-15
tree-sitter: introduce tree-sitter, add nix grammar

inclyc:
Introduce tree-sitter into nixd, for static analysis (#243), AST-based editing & semantic tokens.

Link: https://tree-sitter.github.io/tree-sitter/


inclyc:
nvim tree-sitter integration already has these feature, closing.

For static analysis, I'm thinking about lowering nil ASTs to lower C++ structures, (co-operate with the nil author)

inclyc:
Finally, #257 and #260 . We will not link against nil.

Issue #255, opened 2023-09-10
rename function with default parameters not works properly


luochen1990:
Code to reproduce this issue:

```nix
{a ? 1}: {
    x = a;
}
```

Reproduce steps for 1st usage:

1. move cursor on `a` in the 1st line,  and rename it to `b`
2. trigger the *rename* action,  and found the default text is "a ? 1"
3. if you replace "a ? 1" into "b ? 1", then the second line becomes "x = b ? 1;", which is not as expected.
4. if you replace "a ? 1" into "b", then the second line becomes "x = b;", but the first line becomes "{b}: {", which is not as expected.

Reproduce steps for 2nd usage:

1. move cursor on `a` in the 2nd line,  and rename it to `b`
2. trigger the *rename* action,  and found the default text is "a"
3. if you replace "a" into "b", then the second line becomes "x = b;", but the first line becomes "{b}: {", which is not as expected.

Extra Information:

- using VSCode  extension **Nix IDE v0.2.2**
- $ nixd --version
LLVM (http://llvm.org/):
LLVM version 16.0.6
Optimized build.

inclyc:
This is a known bug caused by limitations from the official nix parser. It focused on the performance of evaluation, and discarded many information for LSPs. 

The possible solutions are:

1. A new parser that is suitable for LSPs. We have experimental tree-sitter-nix branch, but not merged yet.
2. Just link nil with nixd, we have some private chat on Telegram, discussing about linking the two language server with combined abilities (i.e. static analysis from nil, and evaluation & internal inspect from nixd). But the author of nil does not like this, for some reason.

PR #257, opened 2023-09-15
nixd/Syntax: init

inclyc:
This is a major change to our frontend. 

## Motivation

Previously the parser is extracted from NixOS/nix, the official implementation (OI), and it lacks some location information & not easy to do static analysis (because many things are desugared). It is necessary to have an intermediate AST that preserves all information, and it should be able to perform AST transforms that generate `nix::Expr` nodes, which are evaluable.

## The PR

Adds a novel frontend, generated by flex/yacc, with rules copied from OI but structures are rewritten in `nixd::syntax` nodes. These rules has no functional change to OI, but has some language extensions & syntax nodes are preserved, for static analysis & AST-based assistance & provide linting.

## Status

It is working in progress now.

- [x] working parser for basic expressions
- [x] all nixpkgs files.
- [x] tests for above things
- [x] the lowering process to evaluable nodes, i.e. `nix::Expr`, and emit diagnostics, associate between syntax nodes & evaluable nodes
- [ ] used in the nixd server

Link: #255 
Link: #244 
Link: #243


inclyc:
Update: we have a working parser now.

inclyc (commented):
Diagnostic system looks ugly for now

inclyc:
We are doing this in a separated re-usable library: https://github.com/inclyc/libnixf, so closing.

PR #261, opened 2023-09-19
nixd/Server/configuration: use `.nixd.json` for `workspace/configuration` defaults

jonathanjameswatson:
When `nixd` in `workspace/configuration` is unset in VSCode, we get `[null]`. This causes the JSON configuration to be used (not overwritten). However, in Emacs, we get `[{}]` instead, so the JSON configuration is overwritten with the default one. By using the JSON configuration as the defaults for `workspace/configuration`, we get the same behaviour when `workspace/configuration` is unset in VSCode and Emacs. The JSON configuration is no longer ignored when `workspace/configuration` is set, but it never takes priority.

An additional advantage of this change is that when `workspace/configuration` is unset (or set to `null`) while the language server is running, we fall back to the JSON configuration instead of keeping the old deleted `workspace/configuration`.

Feel free to make any improvements to my branch.

### DIFF
--- a/lspserver/include/lspserver/LSPBinder.h
+++ b/lspserver/include/lspserver/LSPBinder.h
@@ -12,8 +12,8 @@ namespace lspserver {
 template <typename T>
 llvm::Expected<T> parseParam(const llvm::json::Value &Raw,
                              llvm::StringRef PayloadName,
-                             llvm::StringRef PayloadKind) {
-  T Result;
+                             llvm::StringRef PayloadKind, T Base = T()) {
+  T Result = Base;
   llvm::json::Path::Root Root;
   if (!fromJSON(Raw, Result, Root)) {
     elog("Failed to decode {0} {1}: {2}", PayloadName, PayloadKind,
--- a/lspserver/include/lspserver/LSPBinder.h
+++ b/lspserver/include/lspserver/LSPBinder.h
@@ -6,6 +6,7 @@
 #include <llvm/ADT/FunctionExtras.h>
 #include <llvm/ADT/StringMap.h>
 #include <llvm/Support/JSON.h>
+
 #include <type_traits>
 
 namespace lspserver {
@@ -13,7 +14,7 @@ namespace lspserver {
 namespace detail {
 
 template <typename T>
-typename std::enable_if<std::is_default_constructible_v<T>, T>::type
+typename std::enable_if_t<std::is_default_constructible_v<T>, T>
 valueOrUninitialized(const std::optional<T> &OptionalDefault) {
   T Result;
   if (OptionalDefault) {
@@ -23,7 +24,7 @@ valueOrUninitialized(const std::optional<T> &OptionalDefault) {
 }
 
 template <typename T>
-typename std::enable_if<!std::is_default_constructible_v<T>, T>::type
+typename std::enable_if_t<!std::is_default_constructible_v<T>, T>
 valueOrUninitialized(const std::optional<T> &OptionalDefault) {
   return OptionalDefault.value();
 }
@@ -54,8 +55,7 @@ llvm::Expected<T> parseParamWithOptionalDefault(
 } // namespace detail
 
 template <typename T>
-typename std::enable_if<std::is_default_constructible_v<T>,
-                        llvm::Expected<T>>::type
+typename std::enable_if_t<std::is_default_constructible_v<T>, llvm::Expected<T>>
 parseParam(const llvm::json::Value &Raw, llvm::StringRef PayloadName,
            llvm::StringRef PayloadKind) {
   return detail::parseParamWithOptionalDefault<T>(Raw, PayloadName,
--- a/lspserver/include/lspserver/LSPBinder.h
+++ b/lspserver/include/lspserver/LSPBinder.h
@@ -6,18 +6,29 @@
 #include <llvm/ADT/FunctionExtras.h>
 #include <llvm/ADT/StringMap.h>
 #include <llvm/Support/JSON.h>
+#include <type_traits>
 
 namespace lspserver {
 
 namespace detail {
+
+template <typename T>
+typename std::enable_if<std::is_default_constructible_v<T>, T>::type
+valueOrDefault(const std::optional<T> &OptionalDefault) {
+  return OptionalDefault ? OptionalDefault.value() : T();
+}
+
+template <typename T>
+typename std::enable_if<!std::is_default_constructible_v<T>, T>::type
+valueOrDefault(const std::optional<T> &OptionalDefault) {
+  return OptionalDefault.value();
+}
+
 template <typename T>
 llvm::Expected<T> parseParamWithOptionalDefault(
     const llvm::json::Value &Raw, llvm::StringRef PayloadName,
     llvm::StringRef PayloadKind, std::optional<T> OptionalDefault = {}) {
-  T Result;
-  if (OptionalDefault) {
-    Result = std::move(OptionalDefault.value());
-  }
+  T Result = valueOrDefault(OptionalDefault);
   llvm::json::Path::Root Root;
   if (!fromJSON(Raw, Result, Root)) {
     elog("Failed to decode {0} {1}: {2}", PayloadName, PayloadKind,
@@ -35,12 +46,14 @@ llvm::Expected<T> parseParamWithOptionalDefault(
   }
   return Result;
 }
+
 } // namespace detail
 
 template <typename T>
-llvm::Expected<T> parseParam(const llvm::json::Value &Raw,
-                             llvm::StringRef PayloadName,
-                             llvm::StringRef PayloadKind) {
+typename std::enable_if<std::is_default_constructible_v<T>,
+                        llvm::Expected<T>>::type
+parseParam(const llvm::json::Value &Raw, llvm::StringRef PayloadName,
+           llvm::StringRef PayloadKind) {
   return detail::parseParamWithOptionalDefault<T>(Raw, PayloadName,
                                                   PayloadKind);
 }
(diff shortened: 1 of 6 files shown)


lspserver/include/lspserver/LSPServer.h
```diff
@@ -66,7 +66,8 @@ class LSPServer : public MessageHandler {
 
   template <class ParamTy, class ResponseTy>
   llvm::unique_function<void(const ParamTy &, Callback<ResponseTy>)>
-  mkOutMethod(llvm::StringRef Method, OutboundPort *O = nullptr) {
+  mkOutMethod(llvm::StringRef Method, OutboundPort *O = nullptr,
+              ResponseTy Base = ResponseTy()) {
```
inclyc, line 70: same as above

inclyc (commented):
Thanks for your contribution!

Left some comment that I think it is unnecessary to change public codes, and also, the commit message should be revised as `path/to/subsystem: change`, with more details about the diff (don't say `fix bug`, instead, say what you have been fixed).

nixd/lib/Server/Controller.cpp
```diff
@@ -257,6 +257,14 @@ Controller::Controller(std::unique_ptr<lspserver::InboundPort> In,
     : LSPServer(std::move(In), std::move(Out)), WaitWorker(WaitWorker),
       ASTMgr(Pool) {
 
+  // JSON Config, run before initialize
+  readJSONConfig();
+  configuration::TopLevel JSONConfigCopy = JSONConfig;
+  updateConfig(std::move(JSONConfigCopy));
+  WorkspaceConfiguration =
```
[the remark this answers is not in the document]
jonathanjameswatson: I have probably overestimated how concurrent the code is, but the reasoning was that the callback to `initialized` could run before the assignment to `WorkspaceConfiguration`, making the resulting `fetchConfig` call fail.

lspserver/include/lspserver/LSPBinder.h
```diff
@@ -12,8 +12,8 @@ namespace lspserver {
 template <typename T>
 llvm::Expected<T> parseParam(const llvm::json::Value &Raw,
                              llvm::StringRef PayloadName,
-                             llvm::StringRef PayloadKind) {
-  T Result;
+                             llvm::StringRef PayloadKind, T Base = T()) {
+  T Result = Base;
```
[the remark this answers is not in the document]
jonathanjameswatson: I would argue that it's reasonable to have an optional argument that defines the defaults for parsing. It is an optional parameter so it shouldn't change the behaviour of the function (although maybe move semantics could be used to make it more efficient).

If I can't change this then I will either have to copy the implementation of this function into Controller.cpp (breaks DRY) or make a more complicated system of defaulting that acts on the parsed `configuration::TopLevel` structs (more error-prone and less efficient). Either of these options would make the code less maintainable. What would you suggest?

lspserver/include/lspserver/LSPServer.h
jonathanjameswatson: I agree that this doesn't make much sense so I will try to move the behaviour to Controller.cpp.

jonathanjameswatson:
I will rebase and squash the commits on my end as they aren't currently intended to be looked at individually.

lspserver/include/lspserver/LSPBinder.h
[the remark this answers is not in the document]
inclyc: > structs (more error-prone and less efficient).

How about use a dedicated function like `updateConfig` and do the work there? I think introduce a new default argument in the basic library is more error prone for large C++ project (since it might implicitly change many things)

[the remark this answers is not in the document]
jonathanjameswatson: I have refactored `parseParam` again so that it no longer takes a default argument and behaves the same way as before (uninitializing `Result` by default or using the default constructor if this is not possible, as well as not compiling if neither is possible). However, it now shares code with `parseParamWithDefault`, which has an argument that determines some default JSON that should be used. I think that this is a reasonable function to add to a general-purpose LSP library.

nixd/lib/Server/Controller.cpp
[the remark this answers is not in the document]
jonathanjameswatson: @inclyc Could you confirm whether or not this could happen?

lspserver/include/lspserver/LSPBinder.h
```diff
@@ -31,6 +50,26 @@ llvm::Expected<T> parseParam(const llvm::json::Value &Raw,
   }
   return Result;
 }
+
+} // namespace detail
+
+template <typename T>
+typename std::enable_if<std::is_default_constructible_v<T>,
```
inclyc, line 57: enable_if_t

nixd/include/nixd/Server/Controller.h
```diff
@@ -41,6 +41,12 @@
 
 namespace nixd {
 
+struct OptionalValue {
```
inclyc, line 44: how about just type-def this optional value?

```
using OptionalValue = std::optional<llvm::json::Value> Value;
```

nixd/lib/Server/Controller.cpp
```diff
@@ -204,9 +205,15 @@ void Controller::fetchConfig() {
         lspserver::ConfigurationParams{
             std::vector<lspserver::ConfigurationItem>{
                 lspserver::ConfigurationItem{.section = "nixd"}}},
-        [this](llvm::Expected<configuration::TopLevel> Response) {
+        [this](llvm::Expected<OptionalValue> Response) {
           if (Response) {
-            updateConfig(std::move(Response.get()));
+            llvm::Expected<configuration::TopLevel> ResponseConfig =
```
inclyc, line 210: Will this actually returns error in case of emacs' reply?

I'm just wondering if we received `[{}]` here why it uses JSONConfig as defaults.

nixd/lib/Server/Controller.cpp
```diff
@@ -233,9 +240,9 @@ void Controller::readJSONConfig(lspserver::PathRef File) noexcept {
     std::ifstream In(File.str(), std::ios::in);
     SS << In.rdbuf();
 
-    if (auto NewConfig = parseConfig(SS.str()))
-      updateConfig(std::move(NewConfig.get()));
-    else {
+    if (auto NewConfig = parseConfig(SS.str())) {
+      JSONConfig = std::move(NewConfig.get());
```
inclyc, line 244: I do think it is a good idea to separate `JSONConfig` from `Config`, but I think maybe you can write a method like 

```
getXXXField(xxx)
```

and if `Config` contains such value returns that one, otherwise returns json config, and the default value as a final fallback.

nixd/include/nixd/Server/Controller.h
```diff
@@ -122,6 +128,7 @@ class Controller : public lspserver::LSPServer {
 
   std::mutex ConfigLock;
 
+  configuration::TopLevel JSONConfig;
   configuration::TopLevel Config; // GUARDED_BY(ConfigLock)
```
inclyc, line 132: Myabe rename this into `LSPWorkspaceConfig`

nixd/include/nixd/Server/Controller.h
jonathanjameswatson: If I did this, then the new `fromJSON` function would apply to all `std::optional<llvm::json::Value>` types across the program, which I don't think is a good idea as I think `std::optional<T>` should instead be used to handle `null` in the general case.

nixd/lib/Server/Controller.cpp
jonathanjameswatson: I have tested this with Emacs and it works without errors. Previously, if the language server received `[{}]`, it would use the defaults of the default constructor, not giving any errors. My changes just make it use a given default (the JSONConfig) instead (line 213).

nixd/include/nixd/Server/Controller.h
jonathanjameswatson: `JSONConfig` is just the config read from `.nixd.json` when the language server starts, and `Config` is the result of `workspace/configuration` with `JSONConfig` used as results. Therefore, I think the names make sense, although maybe `Config` could be changed to a more verbose `LSPWorkspaceWithJSONDefaultsConfig`? However, this may also not be a good name as before `workspace/configuration` is received, `Config` is just set to a copy of `JSONConfig`.

nixd/lib/Server/Controller.cpp
jonathanjameswatson: `Config` already includes defaults from `JSONConfig`, so this shouldn't be necessary. If `Config` was just `workspace/configuration`, then this function would be useful but we would need to have some sort of way to tell whether something is set or not in each component of `configuration::TopLevel`. For example, if `depth` is set to `0` in `Config`, we would not be able to identify if this is just the default value or a number that is explicitly set. Therefore, we would have to change the type of `depth` in `Eval` to `std::optional<int>`, along with making other fields optional. This would reduce totality throughout the code.

nixd/include/nixd/Server/Controller.h
```diff
@@ -122,6 +128,7 @@ class Controller : public lspserver::LSPServer {
 
   std::mutex ConfigLock;
 
+  configuration::TopLevel JSONConfig;
```
inclyc, line 131: maybe you can add some comment here, saying that it is the default value of config

inclyc (approved):

nixd/include/nixd/Server/Controller.h
jonathanjameswatson: On second thoughts, I think I should call it `DefaultConfig` and then add a comment that says it is set to the contents of `.nixd.json` if it exists.

inclyc (approved):
LGTM again

PR #262, opened 2023-09-21
nixd/Sema: add some nixd::Expr node lowering

euphgh:
try implement some nix::Expr node lowering, want to be reviewed.
I suggest rewrite `enum NodeKind`, add `NK_EXPROP` for all op type syntax, this may need change define file.
besides, operator position will not get under current design, like`TODO` in code, can simply ignore?


inclyc (commented):
tests?

inclyc:
Prefer: #263

Issue #266, opened 2023-09-30
It is difficult to debug misconfigured .nixd.json


lf-:
The point of this issue is not so much that my configuration doesn't work, but that it is very hard to figure out why, as I can't seem to find any useful logging of why the evaluation (may have) failed. I have had this issue in the past, which is why I am filing a bug this time to record it as something that's hard.

I am currently working on a project with a `.nixd.json` of the following, and am not getting options completion, but I don't know why:

```json
{
    "options": {
        "enable": true,
        "target": {
            "installable": "",
            "args": ["--impure", "-f", "vm.nix", "nixosConfigurations.boxMicro.options"]
        }
    }
}
```

vm.nix contains the following:

```nix
let flake = builtins.getFlake "/home/jade/co/microvm.nix";
in
{
  nixosConfigurations.boxMicro = flake.inputs.nixpkgs.lib.nixosSystem {
    system = "x86_64-linux";
    modules = [
      flake.nixosModules.microvm
      ./vmcfg.nix
    ];
  };
}
```

 I have set the following options on the server in coc.nvim (I found the log level option by reading the source code):

```json
{
    "languageserver": {
        "nix": {
            "command": "nixd",
            "args": ["-log=debug"],
            "filetypes": ["nix"],
            "settings": {},
        }
}
```

However, the system evaluates, so something is going wrong along the way:

```
co/microvm.nix » nix eval --impure -f vm.nix nixosConfigurations.boxMicro.config.system.build.to
plevel.outPath  
trace: warning: system.stateVersion is not set, defaulting to 23.11. Read why this matters on ht
tps://nixos.org/manual/nixos/stable/options.html#opt-system.stateVersion.
"/nix/store/fgzjygyacafqz9v79hfxizmrk8azdpib-nixos-system-nixos-23.11.20230831.3e52e76"
```

There is a pile of logs here: https://gist.github.com/lf-/1b3b3cc5627b91f8945d8bbe7ba6dca2

Notable parts:

```
E[15:28:05.196] 3364889: Failed to decode workspace/configuration reply: expected object
V[15:28:05.196] 3364889: /* error: expected object */
[
  null
]
V[15:28:05.196] 3364889: <<< {"jsonrpc":"2.0","id":2,"result":[null]}
I[15:28:05.196] 3364889: <-- reply(2)
E[15:28:05.196] 3364889: Failed to decode workspace/configuration reply: expected object
V[15:28:05.196] 3364889: /* error: expected object */
[
  null
]

[...]

PR #267, opened 2023-10-09
nixd/Server/configuration: Log when `workspace/configuration` is `[null]`

jonathanjameswatson:
Previously, when `workspace/configuration` was unset, we would get `[null]`, causing an error to be raised by LLVM. While this error is harmless, it causes confusion (it is mentioned in #266, for example). This PR replaces the error message with a log that mentions `workspace/configuration` is unset. This should be more informative for users using `.nixd.json` instead of `workspace/configuration` to configure nixd.

### DIFF
--- a/nixd/lib/Server/Controller.cpp
+++ b/nixd/lib/Server/Controller.cpp
@@ -207,7 +207,7 @@ void Controller::fetchConfig() {
                 lspserver::ConfigurationItem{.section = "nixd"}}},
         [this](llvm::Expected<OptionalValue> Response) {
           if (Response) {
-            const llvm::json::Value ResponseValue =
+            const llvm::json::Value &ResponseValue =
                 Response.get().Value.value();
             if (ResponseValue.kind() != llvm::json::Value::Array) {
               lspserver::elog(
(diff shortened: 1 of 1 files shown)


nixd/lib/Server/Controller.cpp
```diff
@@ -207,10 +207,16 @@ void Controller::fetchConfig() {
                 lspserver::ConfigurationItem{.section = "nixd"}}},
         [this](llvm::Expected<OptionalValue> Response) {
           if (Response) {
+            const llvm::json::Value FirstConfig =
+                Response.get().Value.value().getAsArray()->front();
```
inclyc, line 211: "getAsArray" might return nullptr, and thus `->front()` will results in SIGSEGV.

nixd/lib/Server/Controller.cpp
```diff
@@ -207,10 +207,24 @@ void Controller::fetchConfig() {
                 lspserver::ConfigurationItem{.section = "nixd"}}},
         [this](llvm::Expected<OptionalValue> Response) {
           if (Response) {
+            const llvm::json::Value ResponseValue =
```
inclyc, line 210:
```suggestion
            const llvm::json::Value &ResponseValue =
```

nixd/lib/Server/Controller.cpp
```diff
@@ -207,10 +207,24 @@ void Controller::fetchConfig() {
...
+                Response.get().Value.value();
+            if (ResponseValue.kind() != llvm::json::Value::Array) {
+              lspserver::elog(
+                  "workspace/configuration response is not an array: {0}",
+                  ResponseValue);
+              return;
+            }
+            const llvm::json::Value FirstConfig =
```
inclyc, line 218:
```suggestion
            const llvm::json::Value &FirstConfig =
```

inclyc (approved):

PR #279, opened 2023-12-22
nixd: use libnixt provided functions

inclyc:


inclyc:
we do not handle errors in libnixt

PR #285, opened 2024-01-02
libnixt: add Kinds.h and Kinds.cpp, to descriminate `nix::Expr`s

inclyc:


libnixt/include/nixt/Kinds.h
```diff
@@ -0,0 +1,40 @@
+/// \file
+/// \brief Determine the kind of nix::Expr.
```
inclyc, line 2: minor fixup: this should be quoted in `, for markdown

PR #287, opened 2024-01-02
libnixt: fix various doc-comments for doxygen

inclyc:


libnixt/include/nixt/PtrPool.h
```diff
@@ -0,0 +1,41 @@
+/// \file
```
inclyc, line 1: this file should not be added.

PR #296, opened 2024-01-14
libnixf: line column range

inclyc:
- [x] fixup test cases


libnixf/include/nixf/Basic/Range.h
```diff
@@ -5,16 +5,29 @@
...
 namespace nixf {
 
-struct OffsetRange {
-  const char *Begin;
-  const char *End;
+struct Point {
+  int64_t Line;
+  int64_t Column;
+  std::size_t Offset;
+};
```
Origami404, line 12: I believe that we should define the meanings of these fields more carefully in the comments here, such as:

- Clearly specify whether a line is terminated by `\r`, `\n`, or `\r\n`. (This also defines where is the begin of a line)
- Clearly define whether 'Column' refers to the character offset from the beginning of the line or the byte offset.
- Clearly define whether 'Offset' refers to the character offset or the byte offset.
- How can we ensure that Line/Column matches the Offset?

Origami404: Or just add a comment: `// See also: Lexer::consume`

libnixf/lib/Parse/RangeBuilder.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "nixf/Basic/Range.h"
+
+#include <cassert>
+#include <stack>
+
+namespace nixf {
+class RangeBuilder {
```
Origami404, line 9: What is this class used for?

Origami404 (approved):

libnixf/include/nixf/Basic/Range.h
inclyc: > I believe that we should define the meanings of these fields more carefully in the comments here, such as:

This should only be created by Lexer, I think. Maybe we can just make the constructor private.

PR #310, opened 2024-01-20
nixd-next: add diagnostic quick fix (code action)

inclyc:


inclyc (commented):
This requires lit-testing

PR #316, opened 2024-01-23
libnixf: use sync tokens for creating unknown nodes (error recovery)

inclyc:


libnixf/src/Parse/Parser.cpp
```diff
@@ -106,6 +119,88 @@ class Parser {
     return LookAheadBuf[N];
   }
 
+  /// \brief Consume tokens until the next sync token.
+  /// \returns The consumed range. If no token is consumed, return nullopt.
+  std::optional<RangeTy> consumeAsUnknown() {
+    Point Begin = peek().begin();
+    bool Consumed = false;
+    for (Token Tok = peek(); Tok.kind() != tok_eof; Tok = peek()) {
+      if (std::find(SyncTokens.begin(), SyncTokens.end(), Tok.kind()) !=
```
inclyc, line 128: just use `SyncTokens.find()` ?

libnixf/src/Parse/Parser.cpp
```diff
@@ -106,6 +119,88 @@ class Parser {
...
+    [[nodiscard]] Diagnostic &diag() const {
+      assert(DiagMissing);
+      return *DiagMissing;
+    }
+  };
+
+  ExpectResult expect(TokenKind Kind) {
+    auto Sync = withSync(Kind);
+    if (Token Tok = peek(); Tok.kind() == Kind) {
+      consume();
```
inclyc, line 177: maybe not `consume` the token will calling `expect`

PR #370, opened 2024-03-29
libnixf: lowering `let .. in ...` binds

inclyc:


libnixf/src/Parse/ParseExpr.cpp
```diff
@@ -253,12 +253,14 @@ std::shared_ptr<ExprLet> Parser::parseExprLet() {
 
   ExpectResult ExpKwIn = expect(tok_kw_in);
 
-  if (!ExpKwIn.ok())
+  if (!ExpKwIn.ok()) {
+    auto Attrs = Act.onExprAttrs(Binds->range(), Binds, Let);
```
inclyc, line 257: `Binds` might be nullptr.

PR #400, opened 2024-04-09
nixd: add basic goto-def support via libnixf (no eval)

inclyc:


nixd/lib/Controller/Definition.cpp
```diff
@@ -0,0 +1,80 @@
...
+    Reply(error("cannot find AST node on given position"));
+    return;
+  }
+
+  const nixf::ParentMapAnalysis *PMA = TU.parentMap();
+  assert(PMA && "ParentMap should not be null as AST is not null");
+
+  const nixf::Node *Var = PMA->upTo(*N, nixf::Node::NK_ExprVar);
+  if (!Var) [[unlikely]] {
+    Reply(error("cannot find AST node on given position"));
```
inclyc, line 35:
```suggestion
    Reply(error("expression is not an variable"));
```

nixd/tools/nixd/test/definition.md
```diff
@@ -0,0 +1,78 @@
+
```
inclyc, line 1: empty

nixd/tools/nixd/test/definition-with.md
```diff
@@ -0,0 +1,79 @@
+
```
inclyc, line 1: empty

PR #411, opened 2024-04-10
nixd: complete nixpkgs

inclyc:


nixd/lib/Controller/Completion.cpp
```diff
@@ -102,11 +163,47 @@ void Controller::onCompletion(const CompletionParams &Params,
...
         Reply(std::move(List));
       }
     }
   };
   boost::asio::post(Pool, std::move(Action));
 }
+
+void Controller::onCompletionItemResolve(const CompletionItem &Params,
+                                         Callback<CompletionItem> Reply) {
+  if (Params.data.empty() || !NixpkgsEval) {
```
inclyc, line 191: move this to the thread pool.

nixd/lib/Controller/Completion.cpp
```diff
@@ -102,11 +163,47 @@ void Controller::onCompletion(const CompletionParams &Params,
...
+    Reply(Params);
+    return;
+  }
+  AttrPathCompleteParams Req;
+  auto EV = llvm::json::parse(Params.data);
+  if (!EV)
+    Reply(EV.takeError());
```
inclyc, line 198: return

PR #430, opened 2024-04-19
nixd: provide package information on hover request

inclyc:


inclyc:
This provides package "hover" information from nixpkgs.

image

PR #450, opened 2024-04-23
editors/nvim: fix missing example folder issue in remote testing

Ruixi-rebirth:


inclyc (commented):
> fix missing test folder issue in remote testing

What is 'remote testing' and what is exactly 'missing test folder' issue?

Ruixi-rebirth:
> > fix missing test folder issue in remote testing
> 
> What is 'remote testing' and what is exactly 'missing test folder' issue?

'remote testing'： the nvim test environment defined by the remote repository
'missing test folder': `./nixd/docs/examples/NixOS_Home-Manager/*`


When I pull the necessary definitions from the remote repository and create the nvim test environment I get the following problem：

image


So I've added a way to handle this: when it can't find the example folder locally, it pulls this test example from the git repo

image

inclyc (approved):
I see. However looks like you are trying to use the flake to setup nixos + home-manager options, why not write `builtins.fetchTarball`-like expressions in the neovim, so thus you don't need to do such "fetching flake to /tmp" ugly stuff.

```
"expr": "builtins.fetchTarball ..." 
```

Ruixi-rebirth:
Oh, I got it.

inclyc:
Note that `fetchTarball` might be relatively slow thus the completion will not pop immediately, which may affect end-user experience.

Ruixi-rebirth:
In terms of how to configure it, I don't think it's necessary to start another devShell on how to write a home-manager configuration. The purpose of the devShells.nvim configuration here is one thing: to experience the beauty of nixd in the easiest way possible for the user.

Although the current way of pulling the examples directory from the repository is ugly, the fact that users can use our examples locally makes it very easy for them to experience the power of `nixd`.

inclyc:
>  the fact that users can use our examples locally

During the first time nixd lauches, the flake inputs will be fetched from the internet, this is rather slow. Run `nix store gc` and then launch the editor environment may reproduce this issue

Ruixi-rebirth:
If so,then we can tell the user when we start the editor that nixd will be slow the first time we start it.

inclyc:
> If so,then we can tell the user when we start the editor that nixd will be slow the first time we start it.

Okay, I agree.

flake.nix
```diff
@@ -50,13 +53,34 @@
...
             nixd
             pkgs.nixpkgs-fmt
+            pkgs.git
             (import ./nixd/docs/editors/nvim-lsp.nix { inherit pkgs; })
           ];
+          inputsFrom = [
+            config.flake-root.devShell
+          ];
           shellHook = ''
+            echo -e "\n\033[1;31mDuring the first time nixd launches, the flake inputs will be fetched from the internet, this is rather slow.\033[0m"
```
Ruixi-rebirth, line 63: I have already added the prompt information

PR #470, opened 2024-04-27
nixd: introduce `--semantic-tokens=true/false` to toggle the feature

inclyc:
Recently @GaetanLepage reported these tokens looks very strange on some color theme, in vim.

Introduce a CLI flag that completely disable this feature at all. Because changing color theme might be non-trivial.


GaetanLepage:
Thank you for this convenient features.
Congratulations for this great nixd major release !

PR #498, opened 2024-05-17
pkg: add a non-generic description

srid:
There are many Nix LSP servers, so let's use a description that's unique to nixd.

Note: this description is shown via LSP in VSCode:

image

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -74,7 +74,7 @@ stdenv.mkDerivation {
 
   meta = {
     mainProgram = "nixd";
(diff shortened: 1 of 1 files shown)


Ruixi-rebirth (approved):

Ruixi-rebirth:
Thanks!

inclyc:
> this description is shown via LSP in VSCode:

FWIW that description is evaluated from nixpkgs, maybe better change the description there.

Issue #499, opened 2024-05-17
nixd dies on shutdown but not in the way eglot expects?


Atemu:
Whenever I shutdown `nixd` via emacs' eglot either via `eglot-shutdown` or by killing the last buffer, eglot complains that the LSP died:

```
(jsonrpc-error "request id=15 failed:" (jsonrpc-error-code . -1) (jsonrpc-error-message . "Server died") (jsonrpc-error-data))
```

Well, of course it died (that's what it's supposed to do) but somehow it doesn't shut down like eglot expects it to?

`haskell-language-server` exits properly from eglot's perspective, so it must be something about nixd.

Here's `nixd --log=debug` stderr from the point I stop it:

```
V[15:06:47.125] 249743: <<< {"jsonrpc":"2.0","id":4,"method":"shutdown","params":null}
I[15:06:47.125] 249743: <-- shutdown(4)
I[15:06:47.126] 249743: --> notify exit
V[15:06:47.126] 249743: >>> {"jsonrpc":"2.0","method":"exit","params":null}
I[15:06:47.158] 249743: --> notify exit
V[15:06:47.158] 249743: >>> {"jsonrpc":"2.0","method":"exit","params":null}
```

Eglot's event buffer I captured in another repro:

```
(:jsonrpc "2.0" :id 5 :method "shutdown" :params nil)
[stderr] V[17:53:05.692] 257822: <<< {"jsonrpc":"2.0","id":5,"method":"shutdown","params":null}
[stderr] I[17:53:05.692] 257822: <-- shutdown(5)
[stderr] I[17:53:05.696] 257822: --> notify exit
[stderr] V[17:53:05.696] 257822: >>> {"jsonrpc":"2.0","method":"exit","params":null}
[stderr] I[17:53:05.731] 257822: --> notify exit
[stderr] V[17:53:05.731] 257822: >>> {"jsonrpc":"2.0","method":"exit","params":null}
[internal] Fri May 17 17:53:05 2024:
(:message "Connection state changed" :change "finished\n")

----------b---y---e---b---y---e----------
```

What I find weird is the two exit calls but I'm not familiar with the LSP.

```
nixd, version: 2.1.2
```

inclyc:
There are merely two problems reported in this issue, the first:

> I[15:06:47.126] 249743: --> notify exit
> V[15:06:47.126] 249743: >>> {"jsonrpc":"2.0","method":"exit","params":null}
> I[15:06:47.158] 249743: --> notify exit
[...]

PR #502, opened 2024-05-18
nixd/Controller: reply `shutdown`

inclyc:


inclyc:
`shutdown` was ignored before this commit. @Atemu could you confirm whether this PR fixes https://github.com/nix-community/nixd/issues/499  ?

Atemu (approved):
Nixd now shuts down "cleanly" from eglot's perspective, thank you!

The exit signal is 9; not sure whether that's expected or not.

inclyc:
> The exit signal is 9; 

What did you mean by "exit signal"?

Atemu:
Sorry, I meant exit code. it's just what eglot reports to me.

inclyc:
> Sorry, I meant exit code. it's just what eglot reports to me. 

No, the exit code should be 0 though. How about paste some logs here then?

nixd/lib/Controller/LifeTime.cpp
```diff
@@ -3,6 +3,7 @@
 /// [Server Lifecycle]:
 /// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lifeCycleMessages
 
+#include "lspserver/Protocol.h"
```
inclyc, line 6: canonicalize this header.

Atemu:
Actually, I have misread: It says "status 9", not "exit code":

```
[eglot] Asking EGLOT (nixfiles/(nix-mode)) politely to terminate
[jsonrpc] Server exited with status 9
```

It's the same status code when I exit haskell-language-server, so it's probably correct.

Getting logs is surprisingly hard because eglot justifyably clears all log buffers upon successful shutdown.

Issue #513, opened 2024-05-28
Wrong suggestions for rec keyword.


voronind-com:
**Describe the bug**
I've got an example where `rec` is actually required, but nixd suggests to remove it.

**Logs (Required)**
[Here](https://git.voronind.com/voronind/nixos/src/commit/82ee67bb358cd58a135af3361994b9b6709f5d90/overlay/Keyd.nix) is an overlay that I made for Keyd.

**To Reproduce**
Check my Nix example.

**Expected behavior**
Nixd sees the `inherit patches` and does not suggest the removal.

**Screenshots**
![image](https://github.com/nix-community/nixd/assets/22127600/ad99ad48-15dc-471c-b9b7-8fbc20a6a8e6)

inclyc:
Yes, looks like indeed a bug.

inclyc:
```nix
rec {
  a = 1;
  b = rec {
    inherit a;
  };
}
```

PR #514, opened 2024-05-29
libnixf/Sema: fix attribute source context of `inherit (expr) ...`

inclyc:
Consider recursive attributes:

    rec {
        inherit (expr) attr;
        inherit attr;
    }

The syntax:

    inherit (expr) attr

is desugared into

    attr = expr.attr

.

The expression `expr.attr` should be evaluated in new environment, created by `rec` keyword

However,

    inherit attr

desugaring to

    attr = attr

should be evaluated in parent environment.

This scoping rule is somehow straightfoward but not really documented. libnixf previously evaluated the second case in "new environment".

Fixes: #513

### DIFF
--- a/libnixf/test/Sema/VariableLookup.cpp
+++ b/libnixf/test/Sema/VariableLookup.cpp
@@ -303,6 +303,9 @@ rec {
   std::shared_ptr<Node> AST = parse(Src, Diags);
   VariableLookupAnalysis VLA(Diags);
   VLA.runOnAST(*AST);
+
+  ASSERT_EQ(Diags.size(), 1);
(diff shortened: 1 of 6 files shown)


libnixf/test/Sema/VariableLookup.cpp
```diff
@@ -290,4 +290,19 @@ TEST_F(VLATest, EmptyLetIn) {
...
+rec {
+  a = 1;
+  b = rec {
+    inherit a;
+  };
+})";
+
+  std::shared_ptr<Node> AST = parse(Src, Diags);
+  VariableLookupAnalysis VLA(Diags);
+  VLA.runOnAST(*AST);
```
inclyc, line 305: this test case is missing asserts.

Issue #517, opened 2024-05-31
support & auto complete nixvim options


inclyc:
Write documentation & serious testing about nixvim modules. It is a nice module system we'd like to cover.

inclyc:
```nix
    programs.nixvim = mkOption {
      default = { };
      type = types.submoduleWith {
        shorthandOnlyDefinesConfig = true;
        specialArgs = {
          hmConfig = config;
          inherit helpers;
        };
        modules = [ (import ./modules/hm.nix { inherit lib; }) ] ++ shared.topLevelModules;
      };
    };
```

currently all nixvim options are nested under "submodules". We have basic support for things like `attrsOf (submoules)` but the options completion stops at `programs.nixvim` so thus it cannot complete the rest.

CC @GaetanLepage

This is a funny case exists in real world, and I think it is mandatory to support it. Let's mark this issue an enhancement for this specific types.

GaetanLepage:
> ```nix
>     programs.nixvim = mkOption {
>       default = { };
>       type = types.submoduleWith {
>         shorthandOnlyDefinesConfig = true;
>         specialArgs = {
>           hmConfig = config;
>           inherit helpers;
>         };
>         modules = [ (import ./modules/hm.nix { inherit lib; }) ] ++ shared.topLevelModules;
>       };
>     };
> ```

Where is this snippet coming from ? Nixvim codebase ?

Anyway, thanks for your efforts in including Nixvim support in nixd !

inclyc:
> Where is this snippet coming from ? Nixvim codebase ?

https://github.com/nix-community/nixvim/blob/d15fade62b743839a20d927d3506d503858f49f0/wrappers/hm.nix#L30

inclyc:
The documentation part see #520

PR #518, opened 2024-06-01
libnixt: mangle `submodule` options to it's declaration

inclyc:
Trying to get completions from nixvim gets trouble, because all of its options are nested under a "submodule".

The PR adds basic support for mangling submodule options into it's declaration, and completion for nixvim works.

Fixes: #517

### DIFF
--- a/libnixt/lib/Value.cpp
+++ b/libnixt/lib/Value.cpp
@@ -27,8 +27,6 @@ bool isSubmodule(nix::EvalState &State, nix::Value &V) {
 
 /// \brief Do proper operations to get options declaration on submodule type.
 nix::Value getSubOptions(nix::EvalState &State, nix::Value &Type) {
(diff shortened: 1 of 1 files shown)


inclyc:
截屏2024-06-01 13 42 19
截屏2024-06-01 13 46 55
截屏2024-06-01 13 47 13

libnixt/lib/Value.cpp
```diff
@@ -6,6 +6,43 @@
...
+  } catch (nix::AttrPathNotFound &) {
+    // The value has no "type" field.
+    return false;
+  }
+  return false;
+}
+
+/// \brief Do proper operations to get options declaration on submodule type.
+nix::Value getSubOptions(nix::EvalState &State, nix::Value &Type) {
+  // Mangle the value if only V.type.name == "submodule"
```
inclyc, line 30: These comments seems to be outdated.

PR #530, opened 2024-06-24
[nixd] Path completion

Origami404:
Support completion for path expression.

image

### DIFF
--- a/libnixf/test/Parse/Lexer.cpp
+++ b/libnixf/test/Parse/Lexer.cpp
@@ -2,7 +2,9 @@
 
 #include "Lexer.h"
 
+#include "Token.h"
 #include "nixf/Basic/Diagnostic.h"
+#include "nixf/Basic/TokenKinds.h"
 
 #include <cstddef>
 
@@ -165,6 +167,18 @@ TEST_F(LexerTest, lexIDPath) {
       tok_id,            // id
       tok_path_fragment, // pa/
       tok_id,            // t
+  };
+  auto Tokens = collect(Lexer, &Lexer::lex);
+  for (size_t I = 0; I < sizeof(Match) / sizeof(TokenKind); I++) {
+    ASSERT_EQ(Tokens[I].kind(), Match[I]);
+  }
+  ASSERT_EQ(Tokens.size(), sizeof(Match) / sizeof(TokenKind));
+}
+
+TEST_F(LexerTest, lexPathStart) {
+  Lexer Lexer(R"(./)", Diags);
+  const TokenKind Match[] = {
+      tok_path_fragment,
   };
   auto Tokens = collect(Lexer, &Lexer::lex);
   for (size_t I = 0; I < sizeof(Match) / sizeof(TokenKind); I++) {
--- a/nixd/lib/Controller/Completion.cpp
+++ b/nixd/lib/Controller/Completion.cpp
@@ -12,6 +12,8 @@
 
 #include <boost/asio/post.hpp>
 
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Path.h>
 #include <semaphore>
 #include <set>
 #include <utility>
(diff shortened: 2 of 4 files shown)


libnixf/test/Parse/Lexer.cpp
```diff
@@ -2,7 +2,9 @@
 
 #include "Lexer.h"
 
+#include "Token.h"
```
inclyc, line 5: move this line near the `#include` above

nixd/lib/Controller/Completion.cpp
```diff
@@ -298,7 +338,23 @@ void Controller::onCompletion(const CompletionParams &Params,
               completeAttrName(Scope, Prefix, Options,
                                ClientCaps.CompletionSnippets, List.items);
             }
-          } else {
+            FinishedCompletion = true;
+          }
+
+          // if we are in a literal path, use PathCompletionProvider only
+          if (!FinishedCompletion) {
```
inclyc, line 345: this boolean check looks ungly, what are we doing here?

What will happen if there is no this condition check?

nixd/lib/Controller/Completion.cpp
```diff
@@ -298,7 +338,23 @@ void Controller::onCompletion(const CompletionParams &Params,
...
+            if (Parent->kind() == Node::NK_ExprPath) {
+              const auto &Path = static_cast<const nixf::ExprPath &>(*Parent);
+              if (Path.parts().isLiteral()) {
+                completeExprPath(File, Path, List.items);
+                FinishedCompletion = true;
+              }
+            }
+          }
+
+          // otherwise, fallback to VLACompletionProvider
```
inclyc, line 356: You can determine whether or not providing VLA completion based on the `upExpr` result. For example if an `ExprPath` is encountered, then provide path completion, otherwise if `ExprVar`, provide VLA completion. Try to determine expr kind first, may make the source code more concise.

inclyc (changes requested):

nixd/lib/Controller/Completion.cpp
Origami404: When writing code, I believe that in future there will be more and more different completion modes added, and ultimately form something like that:

image

IMO the nested levels of codes will be too deep if we just adding `if`, so I refactor this part of code. But since we only have 3 completion modes at this time, I will revert changes here and use nested `if`. 

Another solution here will be some `goto`s.

nixd/lib/Controller/Completion.cpp
```diff
@@ -250,6 +252,43 @@ class OptionCompletionProvider {
...
+    return;
+  }
+
+  llvm::SmallVector<char, 32> Path{PathLiteral.begin(), PathLiteral.end()};
+  if (PathLiteral[0] == '.') {
+    const auto &CurFileDir = path::parent_path(CurFilePath);
+    fs::make_absolute(CurFileDir, Path);
+  }
+
+  if (fs::exists(Path) && fs::is_directory(Path)) {
```
inclyc, line 275: can we do this in std::filesystem?

nixd/lib/Controller/Completion.cpp
inclyc: > IMO the nested levels of codes will be too deep if we just adding if, so I refactor this part of code. But since we only have 3 completion modes at this time,

These three completion modes are `switch`able by the "context" of nix codes. For example, I suppose the attrpath completion only makes sense if there is actually an attrpath user typing, and path completion only makes sense if we are in attrpath.

> Another solution here will be some `goto`s.

No, we are not writing automata.

Aleksanaa:
Also `<foo/bar>` is a path (in `$NIX_PATH`). Of course we can also implement it later

Brisingr05:
To add to the previous comment, path completion was also added to Helix in https://github.com/helix-editor/helix/pull/2608 a few months ago and works OOTB.

PR #549, opened 2024-07-23
nixd/Controller: support "goto definition" for select expressions

inclyc:
Add "goto definition" support for select expressions, e.g.

```nix
with pkgs; [
  nixVersions.nix
#              ^
]
```

```nix
pkgs.stdenv.mkDerivation
#            ^
```

These values are now "Ctrl + Click"-able to nixpkgs locations.

### DIFF
--- a/nixd/lib/Controller/Completion.cpp
+++ b/nixd/lib/Controller/Completion.cpp
@@ -351,17 +351,11 @@ void completeSelect(const nixf::ExprSelect &Select, AttrSetClient &Client,
   // Ask nixpkgs provider to get idioms completion.
   NixpkgsCompletionProvider NCP(Client);
 
-  const auto Handler = [](std::exception &E) noexcept {
-    log(DBG "skipped, reason: {0}", E.what());
-  };
   try {
     Selector Sel = mkSelector(Select, mkIdiomSelector(Var, VLA, PM));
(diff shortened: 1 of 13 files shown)


nixd/lib/Controller/Completion.cpp
```diff
@@ -351,17 +351,11 @@ void completeSelect(const nixf::ExprSelect &Select, AttrSetClient &Client,
...
-  };
   try {
     Selector Sel = mkSelector(Select, mkIdiomSelector(Var, VLA, PM));
     NCP.completePackages(mkParams(Sel, IsComplete), List);
   } catch (IdiomSelectorException &E) {
-    Handler(E);
-    return;
-  } catch (SelectorException &E) {
-    Handler(E);
+    log(DBG "skipped, reason: {0}", E.what());
```
inclyc, line 358: there may be regression for this `catch` because `NoSuchVar` will not be caught.

nixd/lib/Controller/DocumentHighlight.cpp
```diff
@@ -9,6 +9,7 @@
 #include "nixd/Controller/Controller.h"
 
 #include <boost/asio/post.hpp>
+#include <exception>
```
inclyc, line 12: header

inclyc (commented):
LG

Issue #560, opened 2024-07-31
nixf doesn't report unused `@args` with function arguments


Aleksanaa:
**Describe the bug**

`args` in `{ foo, bar }@args` isn't reported to be unused.

**Logs (Required)**

**Configuration**

If you have some custom configuration, please paste it here.

**To Reproduce**
Create a new Nix file and type this:

```
{ foo, bar }@args:

foo + bar
```

`args` is unused. nixd doesn't report this, also neither does nixf:

```
=> nixf-tidy < a.nix 
[]
```

**Expected behavior**
Report a warning on `args` being unused.

**Screenshots**
![image](https://github.com/user-attachments/assets/110aa824-05c9-4f55-b2be-b6b356a57a43)

**Additional context**
Found in https://github.com/nix-community/nixd/issues/558

PR #561, opened 2024-07-31
libnixf/Sema: differentiate lambda argument with and without formals

Aleksanaa:
Fixes https://github.com/nix-community/nixd/issues/560


.gitignore
```diff
@@ -3,3 +3,6 @@
 /html/
 
 __pycache__
+
```
inclyc, line 6: Please
```
echo 'result*' >> .git/info/exclude
```

and revert this file (

inclyc (approved):

inclyc (approved):

Issue #555, opened 2024-07-28
Completion for with pkgs.subpackages


stasjok:
nixd can complete packages in a construction like

```nix
with pkgs; [
    |
[
```

but it can't complete packages for a construction like

```nix
with pkgs.vimPlugins; [
    |
]
```

I think it will be a nice addition and probably not so hard to implement, especially since it already can complete plain `pkgs.vimPlugins.|`.

inclyc:
Yes, we can do.

PR #563, opened 2024-07-31
nixd/Controller: support completion for `pkgs.subpackages`

inclyc:
Fixes: #555

Add auto-completion code-snippet like this:

```nix
    with pkgs.vimPlugins; [
        |
    ]
```

i.e. with a "select" expression.

Screenshot_20240731_211203


inclyc (commented):
TODO: missing tests.

PR #564, opened 2024-08-01
libnixf: change extra-{rec,with},escaping-with diagnostics to warning

Aleksanaa:
Because if you don’t remove such unnecessary patterns in time, you may make subtle mistakes later

### DIFF
--- a/libnixf/src/Basic/diagnostic.py
+++ b/libnixf/src/Basic/diagnostic.py
@@ -192,19 +192,19 @@ diagnostics: list[Diagnostic] = [
     {
         "sname": "sema-extra-rec",
         "cname": "ExtraRecursive",
-        "severity": "Hint",
+        "severity": "Warning",
         "message": "attrset is not necessary to be `rec`ursive",
     },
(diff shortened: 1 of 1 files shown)


libnixf/src/Basic/diagnostic.py
```diff
@@ -192,19 +192,19 @@ class Diagnostic(TypedDict):
...
         "cname": "ExtraWith",
-        "severity": "Hint",
+        "severity": "Warning",
         "message": "unused `with` expression",
     },
     {
         "sname": "sema-escaping-with",
         "cname": "EscapingWith",
-        "severity": "Hint",
+        "severity": "Warning",
```
inclyc, line 207: I think we should keep `sema-escaping-with` to `"Hint"`, this diagnostic seems to be too pedantic.

inclyc, line 207:
```suggestion
        "severity": "Hint",
```

inclyc (approved):
LG

PR #569, opened 2024-08-04
nixd: replace nixpkgs-fmt for nixfmt

sebastienhurtel:
Try to replace nixpkgs-fmt with nixfmt.


sebastienhurtel:
Wrong repo, sorry for the noise!

inclyc:
Nice patch! Why cannot we just accept it?

default.nix
```diff
@@ -8,7 +8,7 @@
   meson,
   ninja,
   nix,
-  nixpkgs-fmt,
+  nixfmt-rfc-style,
```
inclyc, line 11: this dep seems to be a testing quirk that we can remove it. I will push your branch to do that.

flake.nix
```diff
@@ -38,7 +38,7 @@
         };
         regressionDeps = with pkgs; [
           clang-tools
-          nixpkgs-fmt
+          nixfmt-rfc-style
```
inclyc, line 41: same as above

libnixf/default.nix
```diff
@@ -5,7 +5,7 @@
   ninja,
   pkg-config,
   lit,
-  nixpkgs-fmt,
+  nixfmt-rfc-style,
```
inclyc, line 8: this line might be removed

inclyc (commented):
Currently I see nixpkgs-fmt is archived, so I think generally it is a good step to set our default to `nixfmt-rfc-style`. 

Regression tests depending on nixpkgs-fmt might be a funny fake-dependency that could be resolved. See issue https://github.com/nix-community/nixd/issues/570

sebastienhurtel:
Hello,

I assume your talking about [nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt)? 
This project is indeed archived. I tried to remove useless lines as you requested. 
Thank you for your time

nixd/include/nixd/Controller/Configuration.h
```diff
@@ -13,7 +13,7 @@ namespace nixd {
 // NOLINTBEGIN(readability-identifier-naming)
 struct Configuration {
   struct Formatting {
-    std::vector<std::string> command = {"nixpkgs-fmt"};
+    std::vector<std::string> command = {"nixfmt"};
```
inclyc, line 16: Actually only this line will be sufficient if you want to change the default. Some regressions tests might be failing because cannot find *nixfmt*. Maybe wait for #570

inclyc (approved):
LGTM

inclyc:
@Aleksanaa why treefmt is still failing?

inclyc:
Works now.

PR #579, opened 2024-08-10
nixd/Controller: allow setting initial configuration via CLI

inclyc:
Add a CLI flag `-config` used to set initial configuration. For editor clients which does not support workspace configuration, the flag could be used as a fallback.

### DIFF
--- a/nixd/lib/CommandLine/Configuration.cpp
+++ b/nixd/lib/CommandLine/Configuration.cpp
@@ -4,6 +4,7 @@
 #include "nixd/CommandLine/Configuration.h"
 #include "nixd/CommandLine/Options.h"
 #include "nixd/Controller/Configuration.h"
+#include "nixd/Support/JSON.h"
 
 #include "lspserver/Logger.h"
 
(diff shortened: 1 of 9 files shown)


nixd/lib/CommandLine/Configuration.cpp
```diff
@@ -0,0 +1,50 @@
...
+} // namespace
+
+Configuration nixd::parseCLIConfig() {
+  if (DefaultConfigJSON.empty())
+    return {};
+
+  llvm::Expected<llvm::json::Value> V = llvm::json::parse(DefaultConfigJSON);
+
+  if (!V) {
+    llvm::errs()
```
inclyc, line 34: This error seems like something the caller should handle. Reporting the error here would make this function less versatile.

nixd/lib/Controller/LifeTime.cpp
```diff
@@ -183,6 +185,12 @@ void Controller::
       evalExprWithProgress(*Client, getDefaultNixOSOptionsExpr(),
                            "nixos options");
   }
+  try {
+    Config = parseCLIConfig();
+  } catch (LLVMErrorException &Err) {
+    lspserver::elog("{0}, reason: {1}", Err.what(), Err.takeError());
```
inclyc, line 191:
```suggestion
    lspserver::elog("parse CLI config error: {0}, reason: {1}", Err.what(), Err.takeError());
```

PR #580, opened 2024-08-10
nixd/Controller: handle UTF-16 code units for locations

inclyc:
截屏2024-08-11 00 02 30

Handle LSP spec (UTF16) character length for locations.


inclyc (commented):
Parsing the file every time requesting the location may not be a good solution. I'd like to see a caching / preprocessing service for LSP locations.

PR #584, opened 2024-08-17
nixd/Controller: Conditionally include devenv option support by default

k3yss:
This PR adds default support for devenv.sh with the `nixd` LSP, eliminating the need for manual configuration. This enhancement makes `devenv` more intuitive and is likely to promote wider adoption of `nixd`, `devenv`, and `Nix` in general. Specifically this PR does the following things:
1: Fetch `devenv's` option using the `devenv-module-options` from the flake at https://github.com/cachix/devenv
2: Checks whether the devenv executable is present, if it is not it won't include the worker for options. 
3: Loads the LSP in case of `devenv` executable being present.

Requires: https://github.com/cachix/devenv/pull/1390 to be merged

### DIFF
--- a/nixd/lib/Controller/LifeTime.cpp
+++ b/nixd/lib/Controller/LifeTime.cpp
@@ -45,9 +45,10 @@ opt<std::string> DefaultDevenvOptionsExpr{
          "  flake = builtins.getFlake \"github:cachix/devenv\";\n"
          "in\n"
          "  if builtins.hasAttr currentSystem flake.outputs.packages\n"
-         "  then flake.outputs.packages.${currentSystem}.devenv-module-options.options\n"
-         "  else {}")
-};
+         "  then "
+         "flake.outputs.packages.${currentSystem}.devenv-module-options."
+         "options\n"
+         "  else {}")};
 
 opt<bool> EnableSemanticTokens{"semantic-tokens",
                                desc("Enable/Disable semantic tokens"),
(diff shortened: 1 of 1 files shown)


nixd/lib/Controller/LifeTime.cpp
```diff
@@ -34,6 +34,19 @@ opt<std::string> DefaultNixOSOptionsExpr{
          "=  (import <nixpkgs/nixos/modules/module-list.nix>) ++ [ ({...}: { "
          "nixpkgs.hostPlatform = builtins.currentSystem;} ) ] ; })).options")};
 
+opt<std::string> DefaultDevenvOptionsExpr{
+    "devenv-options-expr",
+    desc("Default expression interpreted as devenv option declarations"),
+    cat(NixdCategory),
+    init("let\n"
+         "  currentSystem = builtins.currentSystem;\n"
+         "  flake = builtins.getFlake \"github:cachix/devenv?ref=work/keys/expose-devenv-options\";\n"
```
inclyc, line 43: Connecting to a remote flake may initially take some time due to its initial setup process. Additionally, it's likely that the URL will need to be updated once the changes from this pull request (https://github.com/cachix/devenv/pull/1390) have been merged.

I'd like to [open a request window](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest) here, maybe asking the user whether or not they want to let us fetch it.

inclyc (commented):
Thanks @k3yss, I think it is a good direction to make the while ecosystem user-friendly.

Also, conflicts should be resolved before merging :)

nixd/lib/Controller/LifeTime.cpp
k3yss: @inclyc Thank you for the review, regarding the `open a request window` I don't see that being implemented in `nixd`, is it the case. If not I can try to implement it : )

inclyc: > I don't see that being implemented in `nixd`, is it the case. If not I can try to implement it : )

Not done yet, we can do this in subsequent patches.

PR #598, opened 2024-09-19
docs(editor-setup): Mention built-in in LSP mode

jcs090218:


nixd/docs/editor-setup.md
```diff
@@ -1,4 +1,4 @@
-## We ❤️ Contributions!
+## We ❤ Contributions!
```
inclyc, line 1: why this "heart" should be changed?

nixd/docs/editor-setup.md
```diff
@@ -110,6 +110,11 @@ A simple Emacs Lisp configuration that adds nixd to LSP Mode in the mean time is
                      :server-id 'nixd)))
 ```
 
+> [!NOTE]
+>
+> The client is built-in to lsp-mode after `8.0.0`; you don't need this if you
```
inclyc, line 115: > The client

Which client?

> you don't need this if you

Then what should users do (e.g. modifying the config) in this case? Could you provide *a detailed code snippet* to elaborate?

inclyc:
Hi @jcs090218, why this PR is closed? I think it is a good direction to improve the documentation.

nixd/docs/editor-setup.md
jcs090218: > Which client?

The `lsp-mode` is the client implementation. 🤔 

> Then what should users do (e.g. modifying the config) in this case? Could you provide a detailed code snippet to elaborate?

Users should still use the snippet that is already provided.

---

The note simply informs users that they do not need to customize or add a snippet to their configuration after `8.0.0` since it's built-in; otherwise, it is redundant.

nixd/docs/editor-setup.md
jcs090218: The heart is changed because there is an unrecognized character on Windows.

I've tried to revert it, but nothing has changed. 🤔 (For some reason, it remains the blue heart instead of the original red heart.)

jcs090218:
> Hi @jcs090218, why this PR is closed? I think it is a good direction to improve the documentation.

My goal was to inform the repo owner about the changes. I think I have fulfilled my intention; therefore, I closed it.

I'll reopen it.

PR #613, opened 2024-11-07
flake/deps: nix 2.19 -> 2.24

Mic92:


nixd/lib/Eval/AttrSetProvider.cpp
```diff
@@ -131,24 +132,25 @@ void fillOptionDescription(nix::EvalState &State, nix::Value &V,
...
       State.forceValue(*It->value, It->pos);
 
       // In nixpkgs some examples are nested in "literalExpression"
       if (nixt::checkField(State, *It->value, "_type", "literalExpression")) {
         R.Example = nixt::getFieldString(State, *It->value, "text");
       } else {
         std::ostringstream OS;
-        It->value->print(State.symbols, OS);
+        // TODO: replace this with something that can traverse the symbol table
+        //It->value->print(State.symbols, OS);
```
Mic92, line 153: @inclyc unsure what to do here. We need symbols to be looked up here in the table. I suppose they should come from `IT` somehow?

573:
Looks good to me,

Mic92:
> Looks good to me,

Note that I only made it build, haven't tested anything. And also the TODO above is left to be solved.

nixd/lib/Eval/AttrSetProvider.cpp
inclyc: maybe just:

```C++
It->value->print(State, OS);
```

inclyc (approved):
Thanks @Mic92! I think this is ready to merge?

Mic92:
> Thanks @Mic92! I think this is ready to merge?

Looks like it?

PR #618, opened 2024-11-15
ci: don't run on "pull_request"

inclyc:


.github/workflows/build.yml
```diff
@@ -1,7 +1,7 @@
 name: Meson
 
 on:
-  [push, pull_request]
```
Mic92, line 4: I don't think this is what you want. Because now pull requests from third-party are not build.

I think what you want is:

```
on:
  pull_request:
  push:
    branches:
      - main
```

Issue #507, opened 2024-05-23
autocomplete for builtins


adrian-gierakowski:
**Is your feature request related to a problem? Please describe.**

Would be great to get autocomplete when typing `builtins.` like provided by [nil](https://github.com/oxalica/nil)

inclyc:
> Would be great to get autocomplete when typing `builtins`

I would like to design a system that incrementally evaluate your code on-the-fly, lazily. Not only for `builtins`. I think in this way `nixd` `builtins` documentation will come from *real* nix evaluator, not be hard-coded in the source code (like nil). This is in design review, do you have some suggestions?

adrian-gierakowski:
@inclyc this sounds great! However I guess that will take longer to implement. In the meantime it would be great to have hardcoded completions to keep users from going back to nil 😄

inclyc:
> In the meantime it would be great to have hardcoded completions to keep users from going back to nil

I think you can expect a quick implementation as `builtins` will be relatively trivial, not hardcoded but a *trivial version* of the large-scale eval system.

inclyc:
Ping for this issue. Currently even `lib.*` completion is landed, but NOT `builtins`?

adrian-gierakowski:
Seems like perfect being the enemy of the good kind of situation.

inclyc:
Ping again, I almost forget this...

blindFS:
> Ping for this issue. Currently even `lib.*` completion is landed, but NOT `builtins`?

@inclyc Sorry to bother, `lib.*` completion is not working for me, how to enable it or help debug it?

PR #620, opened 2024-11-19
nixd/tools/nixd-attrset-eval: support completion for "builtins"

inclyc:
Basic support for option completion on the worker process. This is the premise for doing this in LSP.

Link: #507


nixd/include/nixd/Protocol/AttrSet.h
```diff
@@ -31,6 +32,13 @@ constexpr inline std::string_view AttrPathInfo = "attrset/attrpathInfo";
 constexpr inline std::string_view AttrPathComplete = "attrset/attrpathComplete";
 constexpr inline std::string_view OptionInfo = "attrset/optionInfo";
 constexpr inline std::string_view OptionComplete = "attrset/optionComplete";
+
+/// \brief ↩︎ Request information
+constexpr inline std::string_view BuiltinInfo = "attrset/builtinInfo";
```
inclyc, line 37: Also I wonder if we can just use `attrset/attrpathInfo` to complete such logic.

Mic92:
@inclyc did you request my review for something specific or more for a general review?

Mic92:
It looks reasonable to me and similar to what the nix repl does for completing. However I would prefer if @roberth would have a look as well.

libnixt/include/nixt/Value.h
```diff
@@ -70,4 +70,9 @@ selectStringViews(nix::EvalState &State, nix::Value &V,
   return selectSymbols(State, V, toSymbols(State.symbols, AttrPath));
 }
 
+/// \brief Get nix's `builtins` constant
+inline nix::Value &getBuiltins(const nix::EvalState &State) {
+  return *State.baseEnv.values[0];
+}
+
```
roberth, line 77: Nix should provide this, so I've opened https://github.com/NixOS/nix/pull/11914
Feel free to open PRs to fix tech debt like that

nixd/include/nixd/Protocol/AttrSet.h
roberth: I feel like I don't know enough about the protocol to say anything definitive, but primops and module system options seem quite different to me, so I wouldn't be surprised if you wanted to return different fields for them.

libnixt/include/nixt/Value.h
```diff
@@ -70,4 +70,9 @@ selectStringViews(nix::EvalState &State, nix::Value &V,
   return selectSymbols(State, V, toSymbols(State.symbols, AttrPath));
 }
 
+/// \brief Get nix's `builtins` constant
+inline nix::Value &getBuiltins(const nix::EvalState &State) {
+  return *State.baseEnv.values[0];
```
Mic92, line 75: Could do a TODO now that this could use the proper api in the next nix version.

PriceHiller:
Hey! Checking in here.

What all needs to be done to get this working?

More than happy to get the ball rolling assuming I can be of any help.

inclyc:
> Hey! Checking in here.
> 
> What all needs to be done to get this working?
> 
> More than happy to get the ball rolling assuming I can be of any help.

Actually, this PR is related to the technical approach of #625. I’m considering proposing a technical solution that would work well for both #625 and this PR, but I haven’t had the time to write the code recently. If you’re interested, you could explore the connection between the technical approaches of these two PRs and help move this work forward.

inclyc:
Rebased. Resolving merge conflicts.

inclyc:
Closed, prefer https://github.com/nix-community/nixd/pull/677

PR #623, opened 2024-11-23
nixd/lib/Controller: use `CheckReturn` to simplify early-returns (NFC)

inclyc:


nixd/lib/Controller/Completion.cpp
```diff
@@ -376,72 +377,60 @@ void completeSelect(const nixf::ExprSelect &Select, AttrSetClient &Client,
...
-        }());
-      }
-    }
+        }
+      }();
+    }());
   };
   boost::asio::post(Pool, std::move(Action));
 }
+#undef check
```
inclyc, line 433:
```suggestion
```

PR #624, opened 2024-11-23
fix compilation with nix 2.25

Mic92:


inclyc:
> nixt-nightly> ../test/Value.cpp:61:13: error: cannot convert 'NixInt' (aka 'Checked<long long>') to 'long' without a conversion operator

Looks like on macOS (libc++) it is `long long` :cry:

libnixt/test/Value.cpp
```diff
@@ -58,7 +58,7 @@ TEST_F(ValueTest, selectAttrPath) {
   nix::Value &Kern = selectStringViews(*State, Nested, {"c", "d"});
 
   ASSERT_EQ(Kern.type(), nix::ValueType::nInt);
-  ASSERT_EQ(Kern.integer(), 1);
+  ASSERT_EQ(Kern.integer(), nix::NixInt{ 1 });
```
Mic92, line 61: This one should be solved now.

nixd/lib/Eval/AttrSetProvider.cpp
```diff
@@ -54,8 +54,8 @@ std::optional<Location> locationOf(nix::PosTable &PTable, nix::Value &V) {
     return std::nullopt;
 
   Position LPos = {
-      .line = static_cast<int>(NixPos.line - 1),
-      .character = static_cast<int>(NixPos.column - 1),
+      .line = static_cast<long int>(NixPos.line - 1),
```
Mic92, line 57: Do you think macos is still fine? `long int` vs `long long`?

inclyc: Why this type needs to be changed?

nixd/lspserver/include/lspserver/Protocol.h
```diff
@@ -127,12 +127,12 @@ bool fromJSON(const llvm::json::Value &, VersionedTextDocumentIdentifier &,
 
 struct Position {
   /// Line position in a document (zero-based).
-  int line = 0;
+  long int line = 0;
```
inclyc, line 130: How about just keep this line as-is?

nixd/lib/Eval/AttrSetProvider.cpp
Mic92: Will check again. You are probably right.

flake.nix
```diff
@@ -1,6 +1,9 @@
 {
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    # TODO: switch back once nix 2.25 hits nixpkgs-unstable
```
inclyc, line 3: Already merged? https://nixpk.gs/pr-tracker.html?pr=355350

inclyc (approved):

flake.nix
Mic92: Yes. Go ahead.

Mic92: done.

inclyc (approved):

PR #625, opened 2024-11-23
Value doc

inclyc:
This PR introduces `nix-repl`'s `:doc` command to nixd.

Since https://github.com/NixOS/nix/pull/11072, we can utilize the API exposed by `nix::EvalState::getDoc`. (i.e. to provide hover/completions(todo) from nixpkgs's doc comments).

Current status:

- [x] support hovering
- [ ] support completion-list resolve

| | |
|:-:|:-:|
|The `:doc` command in `nix-repl`| Hovering on that value|
|||

### DIFF
--- a/libnixt/include/nixt/Value.h
+++ b/libnixt/include/nixt/Value.h
@@ -70,4 +70,10 @@ selectStringViews(nix::EvalState &State, nix::Value &V,
   return selectSymbols(State, V, toSymbols(State.symbols, AttrPath));
 }
 
+/// TODO: use https://github.com/NixOS/nix/pull/11914 on nix version bump
+/// \brief Get nix's `builtins` constant
+inline nix::Value &getBuiltins(const nix::EvalState &State) {
+  return *State.baseEnv.values[0];
+}
+
 } // namespace nixt
--- a/nixd/include/nixd/Protocol/AttrSet.h
+++ b/nixd/include/nixd/Protocol/AttrSet.h
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <optional>
 #include <string>
 #include <vector>
@@ -31,6 +32,7 @@ constexpr inline std::string_view AttrPathInfo = "attrset/attrpathInfo";
 constexpr inline std::string_view AttrPathComplete = "attrset/attrpathComplete";
 constexpr inline std::string_view OptionInfo = "attrset/optionInfo";
 constexpr inline std::string_view OptionComplete = "attrset/optionComplete";
+
 constexpr inline std::string_view Exit = "exit";
 
 } // namespace rpcMethod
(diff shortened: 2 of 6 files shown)


libnixt/include/nixt/Value.h
```diff
@@ -70,4 +70,10 @@ selectStringViews(nix::EvalState &State, nix::Value &V,
   return selectSymbols(State, V, toSymbols(State.symbols, AttrPath));
 }
 
+/// TODO: use https://github.com/NixOS/nix/pull/11914 on nix version bump
+/// \brief Get nix's `builtins` constant
+inline nix::Value &getBuiltins(const nix::EvalState &State) {
```
inclyc, line 75: maybe remove, not used in this PR

nixd/include/nixd/Protocol/AttrSet.h
```diff
@@ -31,6 +32,7 @@ constexpr inline std::string_view AttrPathInfo = "attrset/attrpathInfo";
 constexpr inline std::string_view AttrPathComplete = "attrset/attrpathComplete";
 constexpr inline std::string_view OptionInfo = "attrset/optionInfo";
 constexpr inline std::string_view OptionComplete = "attrset/optionComplete";
+
```
inclyc, line 35: empty

nixd/lib/Eval/AttrSetProvider.cpp
```diff
@@ -156,6 +157,43 @@ void fillOptionDescription(nix::EvalState &State, nix::Value &V,
...
+      if (Num > MaxItems)
+        break;
+    }
+  }
+  return Names;
+}
+
+std::optional<ValueDescription> describeValue(nix::EvalState &State,
+                                              nix::Value &V) {
+  const auto Doc = State.getDoc(V);
```
inclyc, line 186: `nix::EvalState::getDoc()` is invoked here

nixd/lib/Controller/Hover.cpp
```diff
@@ -87,6 +87,10 @@ class NixpkgsHoverProvider {
     return OS.str();
   }
 
+  static std::string mkValueMarkdown(const ValueDescription &ValueDesc) {
+    return ValueDesc.Doc;
```
inclyc, line 91: here, the documentation is replied directly without any formatting/tokenization.

hsjobeki:
I am also in favor or just returning the content without Rendering it. Maybe as a function on the C-abi. 
Rendering is done using the markdown Extensions from the documentation readme in nixpkgs.

inclyc:
kindly ping @roberth, can we have a structured helper function exposed from nix-expr?

roberth:
Hey @inclyc, sorry for the delay. Happy to help if I can, but it's not quite clear to me what needs to be done in nix-expr.
Currently the inside of the comment is not parsed, except with a markdown parser, of which I'm not even sure that it's the right parser, considering the mismatch with some extended syntax that's used in Nixpkgs.
If, on the other hand, you're looking for some improvements in the interface around the already implemented functionality, we might have some low hanging fruit to pick there.

PR #644, opened 2025-01-13
fix(completion): correctly parse incomplete attrsets and provide completions

Aetherall:
Small DX adjustments:

- Parser distinguishes a little better formals from attrsets
- Completion are triggered when discovering options of an attrset ( no need to try and guess the first letter )

I had to remove a check for the latter, please review carefully !

### DIFF
--- a/libnixf/src/Parse/ParseExpr.cpp
+++ b/libnixf/src/Parse/ParseExpr.cpp
@@ -71,7 +71,7 @@ std::shared_ptr<Expr> Parser::parseExprApp(int Limit) {
 }
 
 std::shared_ptr<Expr> Parser::parseExpr() {
-  // Look ahead 3 tokens.
+  // Look ahead 4 tokens.
   switch (peek().kind()) {
   case tok_id: {
     switch (peek(1).kind()) {
@@ -93,8 +93,15 @@ std::shared_ptr<Expr> Parser::parseExpr() {
       case tok_comma:    // { a ,
       case tok_id:       // { a b
       case tok_ellipsis: // { a ...
-      case tok_r_curly:
         return parseExprLambda();
+      case tok_r_curly:
+        switch (peek(3).kind()) {
+        case tok_colon: // { a } :
+        case tok_at:    // { a } @
+          return parseExprLambda();
+        default:
+          return parseExprAttrs();
+        }
       default:
         break;
       }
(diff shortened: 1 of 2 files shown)


libnixf/include/nixf/Basic/Range.h
```diff
@@ -28,6 +29,10 @@ class Position {
   friend bool operator<=(const Position &LHS, const Position &RHS) {
     return LHS < RHS || LHS == RHS;
   }
+
+  std::string to_string() const {
```
inclyc, line 33: Is this method intended solely for testing purposes? If so, I would prefer to test the position directly.

libnixf/include/nixf/Basic/Range.h
```diff
@@ -123,6 +128,10 @@ class LexerCursorRange {
   [[nodiscard]] PositionRange range() const {
     return {LCur.position(), RCur.position()};
   }
+
+  std::string to_string() const {
```
inclyc, line 132: same as above

nixd/lib/Controller/Completion.cpp
```diff
@@ -386,7 +386,6 @@ void Controller::onCompletion(const CompletionParams &Params,
       const auto AST = CheckDefault(getAST(*TU));
 
       const auto *Desc = AST->descend({Pos, Pos});
-      CheckDefault(Desc && Desc->children().empty());
```
inclyc, line 389: Why was this line removed? I believe removing it might lead to null pointer exceptions at line 390.

inclyc (approved):
This PR looks great to me, and it’s an impressive improvement to the user experience. I just have a few minor suggestions, mostly regarding coding style. Thank you!

libnixf/include/nixf/Basic/Range.h
Aetherall: Yes it is, i will remove it

nixd/lib/Controller/Completion.cpp
Aetherall: Cant remember, its surely a mistake, will revert !

Aetherall:
Thanks for the review !

I would really appreciate also if you could try it out locally, because maybe there are some edge cases not tested I missed out entirely. I never worked on a lexer / parser before.

I will fix the comments after work (3-4h from now in my TZ)

inclyc (approved):
LGTM! I believe this is ready to be merged.

PR #665, opened 2025-03-29
Fix vscodium shell to work like nvim shell.

sterlind:
Unlike the nvim shell, the vscodium shell didn't specify any nixd server settings, so the example `flake.nix` didn't work. Fixed this by:

- Pulling vscodium's user settings.json into a separate file, and filling it with the same contents as the config elaborated in nvim-lsp.nix.

- Moved dev shell generation in the nixd repo's `flake.nix` to a common `mkDevShell` function for both nvim and vscodium (since the `settings.json` is sepecific to the hardcoded `/tmp/NixOS_Home-Manager` path.)

Ideally we'd read the nixd server settings from the same file for both nvim and vscodium, but I don't know Lua so I didn't want to touch it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__
 
 result
 result-*
+.cache
--- a/flake.nix
+++ b/flake.nix
@@ -58,6 +58,43 @@
             hardeningDisable = [ "fortify" ];
           };
           treefmtEval = treefmt-nix.lib.evalModule pkgs ./treefmt.nix;
+          mkDevShell = {cmd, editor}: pkgs.mkShell {
+            nativeBuildInputs = [
+              nixd
+              pkgs.nixfmt-rfc-style
+              pkgs.git
+              editor
+            ];
+            inputsFrom = [ config.flake-root.devShell ];
+            shellHook = ''
+              echo -e "\n\033[1;31mDuring the first time nixd launches, the flake inputs will be fetched from the internet, this is rather slow.\033[0m"
+              echo -e "\033[1;34mEntering the test environment...\033[0m"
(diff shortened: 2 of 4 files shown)


flake.nix
```diff
@@ -58,6 +58,45 @@
...
+            pkgs.mkShell {
+              nativeBuildInputs = [
+                nixd
+                pkgs.nixfmt-rfc-style
+                pkgs.git
+                editor
+              ];
+              inputsFrom = [ config.flake-root.devShell ];
+              shellHook = ''
+                echo -e "\n\033[1;31mDuring the first time nixd launches, the flake inputs will be fetched from the internet, this is rather slow.\033[0m"
```
inclyc, line 72: I have some aversion to these shell scripts; it would be better if the logic could be rewritten in Python.

nixd/docs/editors/vscode-settings.json
```diff
@@ -0,0 +1,28 @@
...
+        "home-manager": {
+          "expr": "(builtins.getFlake \"/tmp/NixOS_Home-Manager\").homeConfigurations.\"user@hostname\".options"
+        },
+        "flake-parts": {
+          "expr": "let flake = builtins.getFlake (\"/tmp/NixOS_Home-Manager\"); in flake.debug.options // flake.currentSystem.options"
+        }
+      }
+    }
+  }
+}
```
inclyc, line 28: add \n at EOF

.gitignore
```diff
@@ -6,3 +6,4 @@ __pycache__
 
 result
 result-*
+.cache
```
inclyc, line 9: why this is necessary?

PR #678, opened 2025-04-27
nixd: enable completion and hover in config sets

### LINKED ISSUE #566: Autocomplete breaks in NixOS module with explicit  config section
**Describe the bug**

If I have a config like:

```
{ lib, pkgs, config, ...}:
{
  systemd.se<TAB>
}
```

autocomplete works

However if I have a config like:

```
{ lib, pkgs, config, ... }: {
  config = {
    systemd.s<TAB>
  };
}

```
auto complete does **not** work

However it does work when I have

```
{ lib, pkgs, config, ... }: {
  config = lib.mkIf true {
    systemd.se<TAB>
  };
}

```

**Logs (Required)**

```
E[16:35:36.158] 22467: option worker reported: -32001: attrname config not found in attrset
```

**Configuration**

If you have some custom configuration, please paste it here.

**To Reproduce**
1. Create a NixOS config like `{ config = {}; }`

**Expected behavior**

**Screenshots**

https://github.com/user-attachments/assets/b94c2457-e2a7-4e6c-bb64-eccc289bb1e3

**Additional context**

not-my-profile:
In NixOS modules options can be defined at the top-level, e.g:

    { foo = true; }

or inside of `config`, e.g:

    { config = { foo = true; }; }

Autocompletion and hover information before only worked for the former because the option resolving code had no special handling for `config`.

Fixes #566.

### DIFF
--- a/nixd/lib/Controller/AST.cpp
+++ b/nixd/lib/Controller/AST.cpp
@@ -315,3 +315,19 @@ nixd::FindAttrPathResult nixd::findAttrPath(const nixf::Node &N,
 
   return R::NotAttrPath;
 }
+
+nixd::FindAttrPathResult
+nixd::findAttrPathForOptions(const nixf::Node &N,
+                             const nixf::ParentMapAnalysis &PM,
+                             std::vector<std::string> &Path) {
+
+  const auto R = findAttrPath(N, PM, Path);
+  if (R == FindAttrPathResult::OK) {
+    // FIXME: Only do this in NixOS module files and in flakes only in module
+    // functions passed to nixpkgs.lib.nixosSystem.
+    if (!Path.empty() && Path[0] == "config") {
+      Path.erase(Path.begin());
+    }
+  }
+  return R;
+}
--- a/nixd/lib/Controller/AST.h
+++ b/nixd/lib/Controller/AST.h
@@ -129,4 +129,12 @@ FindAttrPathResult findAttrPath(const nixf::Node &N,
                                 const nixf::ParentMapAnalysis &PM,
                                 std::vector<std::string> &Path);
 
+/// \brief Heuristically find attrpath suitable for "attrpath" completion.
+/// Strips "config." from the start to support config sections in NixOS modules.
+///
+/// \param[out] Path the attrpath.
+FindAttrPathResult findAttrPathForOptions(const nixf::Node &N,
+                                          const nixf::ParentMapAnalysis &PM,
+                                          std::vector<std::string> &Path);
+
 } // namespace nixd
--- a/nixd/lib/Controller/Completion.cpp
+++ b/nixd/lib/Controller/Completion.cpp
@@ -275,7 +275,7 @@ void completeAttrPath(const Node &N, const ParentMapAnalysis &PM,
                       std::vector<lspserver::CompletionItem> &Items) {
   std::vector<std::string> Scope;
   using PathResult = FindAttrPathResult;
-  auto R = findAttrPath(N, PM, Scope);
+  auto R = findAttrPathForOptions(N, PM, Scope);
   if (R == PathResult::OK) {
     // Construct request.
     std::string Prefix = Scope.back();
--- a/nixd/lib/Controller/Definition.cpp
+++ b/nixd/lib/Controller/Definition.cpp
@@ -240,7 +240,7 @@ Locations defineAttrPath(const Node &N, const ParentMapAnalysis &PM,
                          Controller::OptionMapTy &Options) {
   using PathResult = FindAttrPathResult;
   std::vector<std::string> Scope;
-  auto R = findAttrPath(N, PM, Scope);
+  auto R = findAttrPathForOptions(N, PM, Scope);
   Locations Locs;
   if (R == PathResult::OK) {
     std::lock_guard _(OptionsLock);
--- a/nixd/lib/Controller/Hover.cpp
+++ b/nixd/lib/Controller/Hover.cpp
@@ -147,7 +147,7 @@ void Controller::onHover(const TextDocumentPositionParams &Params,
       }
 
       auto Scope = std::vector<std::string>();
-      const auto R = findAttrPath(N, PM, Scope);
+      const auto R = findAttrPathForOptions(N, PM, Scope);
       if (R == FindAttrPathResult::OK) {
         std::lock_guard _(OptionsLock);
         for (const auto &[_, Client] : Options) {
--- a/nixd/tools/nixd/test/completion/options-in-config.md
+++ b/nixd/tools/nixd/test/completion/options-in-config.md
@@ -1,0 +1,81 @@
+# RUN: nixd --lit-test \
+# RUN: --nixos-options-expr="{ foo.bar = { _type = \"option\"; }; }" \
+# RUN: < %s | FileCheck %s
+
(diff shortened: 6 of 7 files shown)


libnixt/lib/Value.cpp
```diff
@@ -189,6 +189,14 @@ nix::Value nixt::selectOptions(nix::EvalState &State, nix::Value &V,
   }
 
   // Otherwise, simply select it.
-  nix::Value &Nested = selectAttr(State, V, *Begin);
+  nix::Value Nested;
+  try {
+    Nested = selectAttr(State, V, *Begin);
+  } catch (const nix::BaseError &Err) {
```
inclyc, line 195: I think there might be an issue with this technical solution: we would also transform `config.config.foo.config` into `config`, meaning the intermediate `config` would be eliminated as well. I am more inclined to modify [this function](https://github.com/nix-community/nixd/blob/3aa27fde1edcf7b126c70a62aad05d120209363c/nixd/lib/Controller/Completion.cpp#L281). Here, we could add a check to determine whether the first `AttrName` is `config`.

nixd/lib/Controller/Completion.cpp
```diff
@@ -280,6 +280,9 @@ void completeAttrPath(const Node &N, const ParentMapAnalysis &PM,
     // Construct request.
     std::string Prefix = Scope.back();
     Scope.pop_back();
+    if (!Scope.empty() && Scope[0] == "config") {
```
inclyc, line 283: This approach LGTM. However, we also use `findAttrPath` in `Definition.cpp` and `Hover.cpp` to provide go-to-definition & hover docs for options. I kindly ask if you can write a shared function for them, like `findAttrPathForOptions` pand allowing go-to-definition + hovering works inside `config =` consistently?

not-my-profile: How do you feel about doing it in `getNestedAttrPath` rather than introducing a new helper method?

Btw thanks for catching that doing it only in completeAttrPath lost the fix for hover ... I now also added a respective hover test :)

inclyc: > How do you feel about doing it in `getNestedAttrPath` rather than introducing a new helper method?

I think it is a good idea to introduce a new method, because `getNestedAttrPath` is not specific to NixOS module system, and I cannot infer it may omit first `config.` by just reading the function name. So I'd like to prefer a new method.

not-my-profile: That makes sense. While I'm at it I'd also like to improve `findAttrPath` to better deal with function calls. In order to do that I'd like to write some unit tests for it (all the test cases would be testing the same thing and I find lit tests too verbose for this).  However I found out that this isn't that easy with the current design because `findAttrPath` requires some `ParentMapAnalysis` which is obtained via the translation unit but `Controller::getTU` is private. While we could use `friend class` for such unit tests I think it would be nicer to refactor the code so that this isn't necessary? (I'm a bit out of my depth cause I have almost no experience with C++.)

inclyc: > That makes sense. While I'm at it I'd also like to improve `findAttrPath` to better deal with function calls.

Specifically, how do you plan to improve it?

not-my-profile: Currently if you have `foo = someFunc { ... }` then the path inside the attrset is inferred to be the empty path rather than `foo`. Yes we don't know the semantics of the function but not forgetting the prior path would be more correct and work for e.g. `lib.mkIf` and could be made to work with `lib.mkMerge` and other common nixpkgs functions. To still work in Flakes we'd need to implement some heuristics to detect whether a file is a flake or a NixOS module as well as some flake specific semantics ... haven't looked into flake parts yet.

inclyc: > more correct

I don't think correctness can be objectively measured in this context.

> and work for e.g. `lib.mkIf` and could be made to work with `lib.mkMerge` and other common nixpkgs functions.

That said, I understand the point you're making. However, I believe changes like supporting `lib.mkIf`, and potentially `lib.mkMerge` and other common Nixpkgs functions, should be handled in separate PRs.

not-my-profile: Sure, I'm happy to discuss the design in an issue or on Matrix. I'm fine with leaving this to separate PRs. I have introduced the `findAttrPathForOptions` you suggested and added a FIXME comment about my concern that this leads to wrong completions at the top-level of `flake.nix` files.

I ended up improving how documents are specified in tests: #679. If you like this change and merge it before this I could add another test case for a `flake.nix` file with `nixpkgs.lib.nixosSystem` where `config` will also be recognized properly with this PR because the attr path currently resets after a function (I think it would be nice to add a test case for this so that we don't stop getting this right when making future changes).

inclyc (approved):
This approach LGTM, I left some comments just for nit picking.

nixd/lib/Controller/AST.h
```diff
@@ -129,4 +129,12 @@ FindAttrPathResult findAttrPath(const nixf::Node &N,
                                 const nixf::ParentMapAnalysis &PM,
                                 std::vector<std::string> &Path);
 
+/// \brief Heuristically find attrpath suitable for "attrpath" completion.
+/// Strips "config." from the start to support config sections in NixOS modules.
+///
+/// \param[out] Path the attrpath.
```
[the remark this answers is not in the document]
not-my-profile: I copied the documentation from the `findAttrPath` function. If we change it here we should also change it there. However I think returns is wrong? The function returns a `FindAttrPathResult`.

nixd/lib/Controller/AST.cpp
```diff
@@ -315,3 +315,19 @@ nixd::FindAttrPathResult nixd::findAttrPath(const nixf::Node &N,
...
+
+nixd::FindAttrPathResult
+nixd::findAttrPathForOptions(const nixf::Node &N,
+                             const nixf::ParentMapAnalysis &PM,
+                             std::vector<std::string> &Path) {
+
+  const auto R = findAttrPath(N, PM, Path);
+  if (R == FindAttrPathResult::OK) {
+    // FIXME: This is naive because it will wrongly strip away a top-level
+    // "config" in a flake.nix.
```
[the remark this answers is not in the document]
not-my-profile: In a `flake.nix` file top-level keys are `inputs`, `outputs` and `description`. With this function nixd will now complete a top-level `config` which doesn't exist in a `flake.nix` file.

nixd/lib/Controller/AST.h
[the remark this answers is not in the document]
inclyc: I see, I think you are correct.

nixd/lib/Controller/AST.cpp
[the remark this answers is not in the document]
inclyc: > now complete a top-level `config` which doesn't exist in a `flake.nix` file.

Do you mean that we shouldn't do this when a file is not a NixOS module system file? (Not only `flake.nix`)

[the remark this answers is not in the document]
not-my-profile: Yes pretty much although in Flakes we'd still want to do this within module functions passed to `nixpkgs.lib.nixosSystem`. I have clarified the comment :)

inclyc (approved):

PR #679, opened 2025-04-30
nixd/test: specify textDocument/didOpen via nix code blocks

not-my-profile:
This makes the test cases less verbose and enables syntax highlighting but most importantly makes multiline documents readable so that test definitions don't contain lines such as:

    "text":"{\n    x = 1;\n    anonymousLambda = { a }: a;\n    namedLambda = a: a;\n    numbers = 1 + 2.0;\n    bool = true;\n    bool2= false;\n    string = \"1\";\n    string2 = \"${builtins.foo}\";\n    undefined = x;\n    list = [ 1 2 3 ];\n}\n"

Before multiline documents were duplicated in another nix code block (that was ignored by the test runner) for readability but these duplicated code blocks could (and did) of course get out of sync with the actual test code.

### DIFF
--- a/nixd/lspserver/include/lspserver/Connection.h
+++ b/nixd/lspserver/include/lspserver/Connection.h
@@ -40,8 +40,12 @@ public:
 
   JSONStreamStyle StreamStyle = JSONStreamStyle::Standard;
 
+  /// Read one message as specified in the LSP standard.
+  /// A Language Server Protocol message starts with a set of
+  /// HTTP headers, delimited  by \r\n, and terminated by an empty line (\r\n).
   bool readStandardMessage(std::string &JSONString);
 
+  /// Read one message, expecting the input to be one of our Markdown lit-tests.
   bool readDelimitedMessage(std::string &JSONString);
 
   /// \brief Notify the inbound port to close the connection
@@ -51,10 +55,7 @@ public:
               JSONStreamStyle StreamStyle = JSONStreamStyle::Standard)
       : Close(false), In(In), StreamStyle(StreamStyle) {};
 
-  /// Read messages specified in LSP standard, and collect standard json string
-  /// into \p JSONString.
-  /// A Language Server Protocol message starts with a set of
-  /// HTTP headers, delimited  by \r\n, and terminated by an empty line (\r\n).
+  /// Read one LSP message into \p JSONString depending on the configured StreamStyle.
   bool readMessage(std::string &JSONString);
 
   /// Dispatch messages to on{Notify,Call,Reply} ( \p Handlers)
--- a/nixd/lspserver/include/lspserver/Connection.h
+++ b/nixd/lspserver/include/lspserver/Connection.h
@@ -43,10 +43,10 @@ public:
   /// Read one message as specified in the LSP standard.
   /// A Language Server Protocol message starts with a set of
   /// HTTP headers, delimited  by \r\n, and terminated by an empty line (\r\n).
-  bool readStandardMessage(std::string &JSONString);
+  llvm::Expected<llvm::json::Value> readStandardMessage(std::string &Buffer);
 
   /// Read one message, expecting the input to be one of our Markdown lit-tests.
-  bool readDelimitedMessage(std::string &JSONString);
+  llvm::Expected<llvm::json::Value> readDelimitedMessage(std::string &Buffer);
 
   /// \brief Notify the inbound port to close the connection
   void close() { Close = true; }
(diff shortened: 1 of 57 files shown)


nixd/tools/nixd/test/semantic-tokens/basic.md
```diff
@@ -33,24 +32,8 @@
     string2 = "${builtins.foo}";
     undefined = x;
     list = [ 1 2 3 ];
-    null = null;
-    wit = with builtins; [ a b c];
```
not-my-profile, line 37: An example of how the duplicated nix block didn't match the actual document text.

nixd/lspserver/src/Connection.cpp
```diff
@@ -222,24 +223,53 @@ bool InboundPort::readStandardMessage(std::string &JSONString) {
...
+            {"params",
+             {{"textDocument",
+               {
+                   {"uri", NixDocURI.value()},
+                   {"languageId", "nix"},
+                   {"version", 1},
+                   {"text", llvm::StringRef(NixDoc).rtrim()},
+               }}}},
+        };
+        JSONString += DidOpen.dump();
```
inclyc, line 261: Here, it seems that we are generating a JSON file in code and then using another piece of code to parse it, which feels somewhat inefficient. I did take a look locally, and it seems that modifying [nixd/lspserver/src/Connection.cpp#L293](https://github.com/nix-community/nixd/blob/3aa27fde1edcf7b126c70a62aad05d120209363c/nixd/lspserver/src/Connection.cpp#L266) would be a bit tricky. Still, I'd prefer not to have our structured data serialized into a string only to be parsed again by our own program. Do you have any thoughts on this?

nixd/lspserver/src/Connection.cpp
```diff
@@ -222,24 +223,53 @@ bool InboundPort::readStandardMessage(std::string &JSONString) {
 bool InboundPort::readDelimitedMessage(std::string &JSONString) {
   JSONString.clear();
   llvm::SmallString<128> Line;
-  bool IsInputBlock = false;
+  bool IsJSONBlock = false;
```
inclyc, line 226: If we do need to implement this functionality in `readStandardMessage`, I suggest using a **tagged union** (also known as a discriminated union) to achieve it. The purpose of doing this is to avoid the situation where both `IsJSONBlock` and `NixDocURI` are `true` at the same time.

Here's an example of how we could structure it:

```cpp
struct TestMessageState {
    enum class ContextKind { JSONBlock, NixBlock /*, ... */ };

    ContextKind kind;
};
```

inclyc (commented):
The idea looks good to me—I've left some comments regarding the technical approach for implementing it.

nixd/lspserver/src/Connection.cpp
not-my-profile: Good point on parser states being better modeled by an enum than two booleans. I refactored the code accordingly :)

Sidenote: I didn't introduce a tagged union to also contain the NixDocURI because it seems like tagged unions in C++ aren't as ergonomic as the algebraic data types in Rust (just called enums there).

nixd/lspserver/src/Connection.cpp
not-my-profile: This is just test code so I wouldn't worry about the performance. But sure, refactored it to move the JSON parsing into the format-specific read functions. Note that to still log the incoming messages in the main loop they are now serialized again.

nixd/lspserver/include/lspserver/Connection.h
```diff
@@ -40,9 +40,15 @@ class InboundPort {
 
   JSONStreamStyle StreamStyle = JSONStreamStyle::Standard;
 
-  bool readStandardMessage(std::string &JSONString);
+  /// Read one message as specified in the LSP standard.
+  /// A Language Server Protocol message starts with a set of
+  /// HTTP headers, delimited  by \r\n, and terminated by an empty line (\r\n).
+  std::optional<llvm::Expected<llvm::json::Value>>
```
inclyc, line 46: Don't like this double-wrapped json value. I just made some local changes to address this, can you [allow me to edit the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

not-my-profile: turned it on :)

inclyc:
@not-my-profile updated, wdyt about the new commits?

not-my-profile:
Personally I like to avoid modelling expected conditions such as EOF as errors and in Rust I quite often use `Result<Option<T>>` but your change is fine by me. Nice I didn't know about `consume_front` ... I was looking for a method like this with the clangd autocompletion but didn't find it. I agree that `Delimited` should be renamed (and also thought about it before) but think `MarkdownLitTest` might be overly specific ... it suggests that we have different types of lit tests, I'd probably go for `LitTest` (to mirror the `--lit-test` CLI flag). While we're at it we could also rename `JSONStreamStyle` to `InputFormat` and `Standard` to `LSP` (sidenote: Microsoft refers to it as a specification ... not a standard).

Yet another sidenote: I've noticed you tend to merge PRs by squashing the commits ... I do like to separate refactor commits from other commits in the commit history to keep the history as readable as possible. Doing this nicely of course requires interactive rebases ... which I do for my commits.

not-my-profile:
> That's a bigger .rename, unnecessary for this PR?

The Delimited -> LitTest rename is also not strictly necessary for this PR ;) I was just saying if we want to do the other renames as well we could do all the renames at once but I'm fine with splitting them up as well.

PR #680, opened 2025-05-02
build: specify nixVersions.nix_2_28 in default.nix

not-my-profile:
To make it work with NixOS 24.11.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -7,7 +7,7 @@
   llvmPackages,
   meson,
   ninja,
-  nix,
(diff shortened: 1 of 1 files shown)


not-my-profile:
I figured out that I can pass `nix = nixVersions.nix_2_28;` when including `default.nix` via `callPackage` so I'm closing this for now. Changing the parameter to `nixVersions` would make it harder to use a specific nix version which you may want to do.

PR #682, opened 2025-05-02
lspserver: rename `JSONStreamStyle::Delimited` → `LitTest`

inclyc:
Depeneds on: #679


inclyc:
@not-my-profile I cherry-picked original diff as a new PR here, can you review this patch?

not-my-profile:
LGTM. Do we also want to rename JSONStreamStyle to InputFormat and Standard to LSP?

inclyc:
> LGTM. Do we also want to rename JSONStreamStyle to InputFormat and Standard to LSP?

Not very necessary to do IMHO.

PR #709, opened 2025-08-11
flake: update nixpkgs input, nixVersions.nix_2_28 -> nixVersions.nix_2_30

xokdvium:
Nix 2.30 has significanltly improved memory usage and eval performance, which would be nice to have in nixd.


xokdvium:
Huh, UBSAN barfs on:

```
/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-boost-1.87.0-dev/include/boost/asio/impl/thread_pool.ipp:133:7: runtime error: load of value 32, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-boost-1.87.0-dev/include/boost/asio/impl/thread_pool.ipp:133:7 in
```

xokdvium:
Bumping boost seems to help here, but some tests start failing for unclear reasons...

inclyc (commented):
Thanks for your contribution! It would be great if this patch could be merged, but the CI failures need to be fixed first.

xokdvium:
The failures are really weird. When I run lit with a bunch of processes the `regressions/nixd` test suite fails, but that works when running the test suite with `-j1` (passing it to `lit`).
This is cursed. I suppose something in nixpkgs broke this.

xokdvium:
Ok, so the boost thing is broken because it really does have an uninitialized variable: https://github.com/boostorg/asio/commit/30b5974ed34bfa321d268b3135ffaffcb261461a

inclyc (approved):

inclyc:
LGTM, thank you!

PR #719, opened 2025-10-01
feat: Use nixpkgs to resolve `with` liveness

booxter:
Since 7683220f we always mark nested `with` expressions alive. This is
technically sound because, indeed, there's no general way to determine
whether a particular attribute comes from one of the nested `with`s.

Sadly, this change made it impossible to detect common problematic
patterns in nixpkgs tree like:

```
meta = with lib; {
  maintainers = with lib.maintainers; [ ... ];
};
```

This would be very useful in our ongoing attempt to automate cleanup of
these patterns from the whole `nixpkgs` tree, see:

https://github.com/technowledgy/refactor-tractor/pull/14

This change attempts to address this problem. It is observed that nixd
completion controllers already use nixpkgs attrset and heuristics to
look for known attributes in `pkgs` and `lib`, assuming these belong to
nixpkgs. We can do the same when analysing `with` scopes.

1. This patch reuses the existing nixpkgs client infrastructure for
   `Sema/VariableLookup.cpp`. Since variable analysis is part of
   `libnixf` while nixpkgs client and LSP are part of `nixd`, we have to
   relocate a number of modules plus the `nixd-attrset-eval` tool
   outside the `nixd` directory under `common`.

2. The `nixd-attrset-eval` tool was truncating the number of returned
   values by 30, which is useful for interactive use in completion
   helper context, but is not enough for comprehensive variable scope
   analysis. For this reason, the json protocol for the tool was
   extended to support `MaxItems` attribute that can be set to `0` to
   indicate that no truncation should occur.

3. On performance optimizations: The new nixpkgs client is initialized
   once per call to `VLA`. It will load and cache discovered attributes
   as needed for the `with` resolution procedure. (In the future, it may
   be used for other needs as needed.)

   Note: It could be beneficial to allow reuse of the same nixpkgs
   client with its cached results across multiple calls to `VLA`. This
   is left for a future exercise, if and when consuming tools like
   `nixd-tidy` gain support for multi-file operation.

4. On build system: Because the project supports "global" meson build
   from the root directory as well as component specific builds for
   `libnixf` and `nixd`, we have to setup scaffolding to reuse the
   `common` library in both. This is achieved with subprojects and a
   bunch of symlinks.

5. Because `libnixf` / `nixd-tiny` can be used interactively, and each
   call to `nixd-attrset-eval` is logged in extreme details (including
   complete json protocol comms between the tool and its driver), we
   have to suppress this output not to pollute the caller's stdout with
   these debug messages. Some adjustments were done to allow to disable
   these log messages when the tool is called from `VLA` nixpkgs client.

6. VLA test timeout is bumped from 30s to 120s because it now has to run
       nixd-attrset-eval. The Sema test now takes time comparable to nixd
       regression suite for this reason. 120s is the timeout taken from nixd
       regression suite.


booxter:
@wolfgangwalther FYI. I tried to capture the design of the feature in the commit message in detail.

I still need to fix some compilation that are, apparently, now caught because I moved some from nixd to libnixf that - I assume - has slightly different compilation regime (?)

Full disclosure: I used Sonnet LLM to bridge the gap between my lack of Meson (and to a lesser extent - C++) knowledge. I am not sure if the approach taken here in regards to cross-linking multiple "subprojects" to a "common" subproject is solid. If someone can point me to a better way of achieving the same, please let me know.

common/lib/Eval/AttrSetProvider.cpp
```diff
@@ -162,22 +162,22 @@ void fillOptionDescription(nix::EvalState &State, nix::Value &V,
...
+                                       std::string_view Prefix, int Limit) {
   std::vector<std::string> Names;
+  const bool Unlimited = Limit <= 0;
 
   // FIXME: we may want to use "Trie" to speedup the string searching.
   // However as my (roughtly) profiling the critical in this loop is
   // evaluating package details.
   // "Trie"s may not beneficial because it cannot speedup eval.
   for (const auto *AttrPtr : Scope.attrs()->lexicographicOrder(State.symbols)) {
+    // Dump to file for debugging.
```
booxter, line 174: remove; debugging attempts remnants

common/lib/Support/ForkPiped.cpp
```diff
@@ -40,6 +40,6 @@ int nixd::forkPiped(int &In, int &Out, int &Err) {
 
   In = PipeIn[WRITE];
   Out = PipeOut[READ];
-  Err = PipeOut[READ];
+  Err = PipeErr[READ];
```
booxter, line 43: double-check if it's still needed

libnixf/src/Sema/VariableLookup.cpp
```diff
@@ -74,6 +85,23 @@ void VariableLookupAnalysis::emitEnvLivenessWarning(
   }
 }
 
+static constexpr std::array<std::string_view, 2> KnownIdioms = {
+    "pkgs",
```
booxter, line 89: To make `pkgs` work, we'll need to assume `pkgs` == `import nixpkgs {}` below and specially handle it.

libnixf/src/Sema/VariableLookup.cpp
```diff
@@ -105,6 +134,30 @@ void VariableLookupAnalysis::lookupVar(const ExprVar &Var,
     for (const auto *WithEnv : WithEnvs) {
       Def = WithDefs.at(WithEnv->syntax());
       Def->usedBy(Var);
+
+      // Attempt to resolve the variable with nixpkgs client.
+      bool IsKnown = false;
+      const auto &With = static_cast<const ExprWith &>(*WithEnv->syntax());
+
+      auto FullPath = With.fullPath();
```
booxter, line 142: I construct path string here, then split it. Maybe I should return a list of attr path elements instead and avoid messing with full path string.

libnixf/test/Sema/VariableLookup.cpp
```diff
@@ -350,4 +350,36 @@ __curPos
   ASSERT_EQ(Diags.size(), 0);
 }
 
+TEST_F(VLATest, NestedWith_KnownMaintainer) {
```
booxter, line 353: to add more test cases; incl. for `pkgs`.

nixd/lspserver/include/lspserver/LSPServer.h
```diff
@@ -47,7 +49,8 @@ class LSPServer : public MessageHandler {
   void callMethod(llvm::StringRef Method, llvm::json::Value Params,
                   Callback<llvm::json::Value> CB, OutboundPort *O) {
     llvm::json::Value ID(bindReply(std::move(CB)));
-    log("--> call {0}({1})", Method, ID.getAsInteger());
```
booxter, line 50: to consider introducing a `maybe_log` or setting log level such that `log` becomes a no-op.

common/lib/Support/ForkPiped.cpp
booxter: Confirmed it's not required. Dropping.

booxter: @inclyc though this change doesn't belong to this PR, I wonder if it's still a legit issue to patch. Thoughts?

inclyc (commented):
Overall, I believe this PR fully addresses the issues encountered in [the refactoring scenario](https://github.com/technowledgy/refactor-tractor/pull/14). **And I agree that we should do this.** However, this patch is relatively large in scale and involves significant changes to the overall structure. Below are some of my thoughts:

> 1. This patch reuses the existing nixpkgs client infrastructure for `Sema/VariableLookup.cpp`. Since variable analysis is part of `libnixf` while nixpkgs client and LSP are part of `nixd`, we have to relocate a number of modules plus the `nixd-attrset-eval` tool outside the `nixd` directory under `common`.

When I initially designed libnixf, its "product positioning" was to be a "foundational library with minimal dependencies" (only relying on header-only libraries) and with runtime dependencies limited to the standard library. I made this decision because I hoped this library would ultimately be maximally reusable, for example:

1. Writing an independent linter based on this library
2. Replacing the official NixOS/nix parser with libnixf
3. Enabling the development of an LSP based on this library
4. Providing language bindings like [Python](https://github.com/inclyc/pynixf)

However, the current design introduces full LSP-required dependencies into libnixf, which might be inappropriate.

> 4. On build system: Because the project supports "global" meson build from the root directory as well as component specific builds for `libnixf` and `nixd`, we have to setup scaffolding to reuse the `common` library in both. This is achieved with subprojects and a bunch of symlinks.

I'm not very fond of meson's "subprojects" feature, as it forces all projects to have a hard-coded "subprojects" directory. Emotionally, I don't want this project to become like that.

I haven't carefully reviewed the specific code implementation yet, but perhaps we can first discuss how to technically solve the "with" liveness issue while keeping libnixf from introducing excessive dependencies?

inclyc (commented):
I just took a brief look at this implementation, and I think its overall correctness should be fine.

booxter:
I would have to double check but I think it will use `<nixpkgs>`. I don't think `nixd` encloses nixpkgs lib on build. @wolfgangwalther 

I may give an answer in a day or two if you don't find out yourself before then.

wolfgangwalther:
> I would have to double check but I think it will use `<nixpkgs>`.

Ah, that makes sense. Annoying, but likely. Will update my system to make sure I get all the latest data.

wolfgangwalther:
> Wonder if we should pin nixpkgs for the refactor tractor using NIX_PATH so that it's pure in this way.

This works in principle, but one needs to be careful - it needs a second checkout with the same base commit, because (I assume) it ends in some cycle of reloading a new changes nixpkgs after every modification the script makes, otherwise...

booxter:
It was used in refactor-tractor only; which is now archived. Closing because we don't care anymore if GitHub harvests the commits.

PR #721, opened 2025-10-06
libnixf: check `builtins.` prefix from variable lookups

inclyc:


philiptaron:
> This information is obtained directly from the C++ library and is not encoded in the nixd code.

Wonderful.

MattSturgeon:
> This information is obtained directly from the C++ library and is not encoded in the nixd code.

My only concern with this approach is when using this for nixpkgs specifically, is nixpkgs is supposed to eval on multiple implementations & versions of nix.

E.g. if a new version of nix adds or remove something from the prelude, nixpkgs wouldn't want to change its linting rules until that change was reflected in all "supported" nix versions (i.e. when we bump `minFeatures` to include the change).

I wonder if there is a way to configure this linting and use the C++ information merely as the default? E.g. we could have an `--enforce-globals` list:
```sh
--enforce-global true \
--enforce-global false \
--enforce-global map \
--enforce-global throw \
# ...
```
When used, only the explicitly listed "globals" would be treated.

IDK if such an interface would need a way to both append to the default list or override it? For nixpkgs we'd want to override the list, but maybe others would want a different behaviour?

philiptaron:
I would be _shocked_ if Nix introduced a new name to the prelude; they've almost always been added to `builtins`.

trueNAHO:
> > I would be _shocked_ if Nix introduced a new name to the prelude; they've almost always been added to `builtins`.
> 
> I don't know how comfortable I am relying on that assumption, when the result of something being added/moved/removed would result in linting errors across nixpkgs _and_ fixing those lint errors would break eval on older still-supported versions (and other supported implementations).
> 
> I guess the questions are 1) when did the prelude last change? 2) are there any plans to move things into or out of the prelude? That would give us _an idea_ of how unreasonably my discomfort is 😁

That is more or less what https://github.com/NixOS/nixpkgs/pull/449123#issuecomment-3375873224 is trying to figure out.

inclyc:
> To avoid problems with later nix versions adding new globals - how about we achieve consistency by _enforcing_ `builtins.` prefixes to _all_ of these?
> 
> This does not suffer from the same problem. It is also much better to feature detect for `lib/minfeatures.nix` in Nixpkgs.
> 
> Ultimately, we just want consistency and nothing to nit-pick about, right?

Too pedantic to write `builtins.true` ?

eclairevoyant:
> I guess the questions are 1) when did the prelude last change?

Every globally-exposed primop has been present since nix 2.3, except for `break` (added 3 years ago in https://github.com/NixOS/nix/pull/5416), which is present in all versions since nix 2.9.

MattSturgeon:
I understand if this is considered not worth it, currently. But in an ideal world, I'd like to see `--exclude-prelude` and `--include-prelude`.

`--exclude-prelude` would allow making specific exceptions to the rule. E.g. `--exclude-prelude fromTOML` would enforce `builtins.fromTOML` instead of `fromTOML` for consistency with `builtins.fromJSON`.

`--include-prelude` would allow overriding the rule with an explicit list. E.g. `--include-prelude true --include-prelude false` would act as if `true` and `false` are the only prelude items. When used, there would be no need to check nix's actual preludes.

While I think this would be useful beyond nixpkgs, it would _also_ allow nixpkgs to handle backwards compatibility _if_ anything does get added to the prelude in the future. Based on @eclairevoyant's comment this seems infrequent and unlikely, but still possible.

Personally, I think `--exclude-prelude` seems more useful generally. `--include-prelude` is kinda niche and, I assume, more effort to implement. It'd only be useful for projects that want to use a strict subset of the prelude.

eclairevoyant:
Side note: "prelude" isn't a documented term anywhere in nix, I think it's a haskell thing.
(Though, not trying to get too in the weeds on naming, before deciding if it's even worth it obviously.)

Issue #747, opened 2025-12-29
Wrong warning in nixpkgs repo


qweered:
For all packages with fetchurl i see this warning

`overriding a builtin name `fetchurl` is discouraged, rename it to avoid confusionnixf(sema-primop-overridden)`

<img width="891" height="166" alt="Image" src="https://github.com/user-attachments/assets/7845405c-4f89-43c4-a1d4-627fd25d8c7d" />

PR #750, opened 2025-12-29
libnixf/Sema: only flag warnings if user try to override "prelude" builtins

inclyc:
Accept builtins that are namespaced (e.g., builtins.path), but overriding "prelude"-builtins (e.g., `map`, `__lessThan`) should still warn.

Fixes: https://github.com/nix-community/nixd/issues/747
Related: https://github.com/nix-community/nixvim/pull/4121

CC @khaneliman, @MattSturgeon, @wolfgangwalther


inclyc:
Merged! Appreciate the detailed comparison between nixvim and nixpkgs.

The strange auto-fix behavior (`inherit (builtins) map` -> `inherit () map`) you noticed is indeed interesting—since it's reproducible on the latest main/unstable, let's track that as a standalone bug to avoid cluttering this change. Thanks again for the help!

PR #784, opened 2026-02-09
nixd/Completion: fix isIncomplete flag handling

Mr-Stoneman:
When typing a attribute like services.x only the first 30 solutions would be displayed and the isIncomplete flag was always set to false. Moving the bounds check fixed this and isincomplete can now be true.

### DIFF
--- a/nixd/lib/Controller/Completion.cpp
+++ b/nixd/lib/Controller/Completion.cpp
@@ -42,10 +42,10 @@ struct ExceedSizeError : std::exception {
 };
 
 void addItem(std::vector<CompletionItem> &Items, CompletionItem Item) {
+  Items.emplace_back(std::move(Item));
   if (Items.size() >= MaxCompletionSize) {
     throw ExceedSizeError();
   }
-  Items.emplace_back(std::move(Item));
 }
 
 bool hasConcreteChild(const Node &N) {
--- a/nixd/tools/nixd/test/completion/options-incomplete.md
+++ b/nixd/tools/nixd/test/completion/options-incomplete.md
@@ -62,7 +62,7 @@
      CHECK: "id": 1,
 CHECK-NEXT:  "jsonrpc": "2.0",
 CHECK-NEXT:  "result": {
-CHECK-NEXT:    "isIncomplete": false,
+CHECK-NEXT:    "isIncomplete": true,
 ```
 
 


inclyc (commented):
I don't understand how this fixes the issue. Also, regression tests should be added to ensure that the fix is verified.

Mr-Stoneman:
This fixes the issue by checking the size of the vector after adding the last element. What was happening before this was that, there were only 30 Items, because the size was checked before adding a element the max size of Items was 29 and so never hit the 30 need to throw the error.

Mr-Stoneman:
I ran the regression test locally, they all passed. I do not know how to add them here sry.

Mr-Stoneman:
Ty I was struggling with that.

inclyc (commented):
The tests passed anyway however I think the real problem here is `nixd/lib/Eval/AttrSetProvider.cpp` where the options list is truncated but it does not notify the Controller the list is truncated.

https://github.com/nix-community/nixd/blob/a6dfbbe41eaa9561fbbb0ece336fc7771810b183/nixd/lib/Eval/AttrSetProvider.cpp#L366-L368

inclyc:
> Regarding the unconditionally growth, the `AttrSetProvider.cpp` file also has a `constexpr int MaxItems = 30;` which also impacts the behavior of this bug. As i currently understand it this `Maxitems = 30` limits the available `Names`, that are used to fill the `Items`

So maybe the correct fix is to add a `isIncomplete` field to `AttrSetProvider.cpp` though, the Controller is notified that the list it receives is also incomplete, HDYT?

Mr-Stoneman:
> The tests passed anyway however I think the real problem here is `nixd/lib/Eval/AttrSetProvider.cpp` where the options list is truncated but it does not notify the Controller the list is truncated.
> 
> https://github.com/nix-community/nixd/blob/a6dfbbe41eaa9561fbbb0ece336fc7771810b183/nixd/lib/Eval/AttrSetProvider.cpp#L366-L368

Hey @inclyc I recently looked at this test again. I think the reason that it passes, is that it tests if the flag is set to `false`. But if I understand the flag correctly it should be `true` because there are more than 30 `Items` (40).

https://github.com/nix-community/nixd/blob/7c2a68af8abf5b6598a28921abc9c6d5bf2b5323/nixd/tools/nixd/test/completion/options-incomplete.md?plain=1#L2
https://github.com/nix-community/nixd/blob/7c2a68af8abf5b6598a28921abc9c6d5bf2b5323/nixd/tools/nixd/test/completion/options-incomplete.md?plain=1#L65

inclyc:
Just to be clear: I agree with your assessment that this isn't the right spot, but I think our underlying approach is wrong. We should treat the list as truncated if *any* `AttrSetProvider` says so.

> The tests pass, but the actual problem lives in `nixd/lib/Eval/AttrSetProvider.cpp` — truncation happens there without notifying the Controller.
> 
> https://github.com/nix-community/nixd/blob/a6dfbbe41eaa9561fbbb0ece336fc7771810b183/nixd/lib/Eval/AttrSetProvider.cpp#L366-L368.

The current fix works, but it feels like a band-aid rather than a proper solution.

Mr-Stoneman:
Sry my bad should have read through the whole thread when looking at this again. 
1. How would you like to return the `isIncomplet` flag? Just as part of the Response? 
2. Should all `AttrSetProvider` return such a flag? Or just the ones that already check the length?

https://github.com/nix-community/nixd/blob/a6dfbbe41eaa9561fbbb0ece336fc7771810b183/nixd/lib/Eval/AttrSetProvider.cpp#L179-L181

inclyc:
I have re-implemented this in https://github.com/nix-community/nixd/pull/887 and I think this could be closed.

Mr-Stoneman:
Nice, your fix looks quite a bit more robust than mine. Thank you  for implementing it.

PR #755, opened 2026-01-03
[DRAFT] feat(code-action): add refactoring and quickfix actions for Nix code

takeokunn:
## Summary

This PR implements comprehensive code actions for nixd, addressing #466 and incorporating community suggestions from the discussion thread.

**Note: This is a draft PR. Features will be split into separate PRs for incremental review and merge.**

### Quick Fixes (diagnostic-based)

- **Add undefined variable to lambda formals** - `{x}: y` → `{x, y}: y`
- **Remove unused lambda formal** - `{x, y}: x` → `{x}: x`
- **Remove unused let binding** - `let unused = 1; in 2` → `2`

### Refactoring Actions

- **Quote/Unquote attribute name** - `foo` ↔ `"foo"`
- **Flatten attribute set** - `{ foo = { bar = 1; }; }` → `{ foo.bar = 1; }`
- **Pack attribute set** - `{ foo.bar = 1; }` → `{ foo = { bar = 1; }; }`
- **Convert JSON to Nix** - Selection-based JSON → Nix conversion
- **Open in noogle.dev** - Quick link to `lib.*` function documentation
- **Extract expression to file** - Command-based extraction (client handles file creation)
- **Convert `with` to `let/inherit`** - `with lib; { x = foo; }` → `let inherit (lib) foo; in { x = foo; }`
- **Convert to inherit** - `{ x = x; }` → `{ inherit x; }` / `{ a = b.a; }` → `{ inherit (b) a; }`
- **Convert to explicit binding** - `{ inherit x; }` → `{ x = x; }`
- **Rewrite string** - `"hello"` ↔ `''hello''`

### LSP Enhancements

- **Hierarchical CodeActionKinds** - `refactor.extract`, `refactor.rewrite`, `source.fixAll`
- **isPreferred marking** - Auto-fix prioritization for specific diagnostics
- **source.fixAll** - Apply all non-conflicting quickfixes at once

## Changes

| File                                    | Changes                               |
|-----------------------------------------|---------------------------------------|
| `libnixf/src/Sema/VariableLookup.cpp`   | +127 lines - Add fixes to diagnostics |
| `nixd/lib/Controller/CodeAction.cpp`    | +700 lines - Refactoring actions      |
| `nixd/lspserver/include/lspserver/Protocol.h` | Kind constants              |
| `nixd/lspserver/src/Protocol.cpp`       | Kind definitions                      |
| `nixd/lib/Controller/LifeTime.cpp`      | Server capabilities                   |
[...]

Issue #466, opened 2024-04-26
nixd: code actions


inclyc:
- [ ] add undefined variable as top-level formal
- [ ] split an subexpression, with `import ./file`
- [ ] `with ..; `  let / inherit

VTimofeenko:
+

- [ ] flatten attrset:
 
    ```nix
    { foo = { bar = "foz"; } };
    ```
    to
    ```nix
    { foo.bar = "foz"; }
    ```
- [ ] pack into nested attrset:

    ```nix
    foo.bar = "foz";
    foo.baz = "faz";
    ```
    to
    ```nix
    foo = { 
      bar = "foz";
      baz = "faz";
    };
    ```

inclyc:
Thanks, any more ideas are welcomed!

zspher:
from: https://github.com/oxalica/nil/blob/c8e8ce72442a164d89d3fdeaae0bcc405f8c015a/docs/code_actions.md
- [ ] `add_to_top_level_lambda_param`
Add an undefined name to the top-level lambda.

```nix
{ foo }: foo + bar
```
=>
```nix
{ foo, bar }: foo + bar
```
- [ ] `quote_attr` and `unquote_attr`
Rewrite between attribute names and double quoted strings
```nix
{ foo = bar; }
```
<=>
```nix
{ "foo" = bar; }
```

inclyc:
> Is this planned or in development? No pressure, just wanted to know since I currently have to switch LSPs when I need any of these. In addition one suggestion is converting json to nix which is quite common (think waybar).

It is something I plan to pursue, but currently, I don’t have enough bandwidth to move it forward. If you're particularly interested in these features, you might consider supporting the project financially. That said, this doesn’t mean I won’t develop them without funding—I’m just not able to prioritize them at the moment.

inclyc:
Almost done.

0xda157:
adding a remove `with` code action would be nice. something like `with pkgs; [ nixd ]` --> `[ pkgs.nixd ]`

Malix-Labs:
@0xda157 does the other way around exists?

0xda157:
> [0xda157] does the other way around exists?

It doesn't exist in nixd currently as far as can tell

PR #785, opened 2026-02-14
feat(code-action): add quickfix to remove extra rec keyword

takeokunn:
## Summary

Add a code action that removes unnecessary `rec` keywords from attribute sets, marked as a preferred quickfix.

**Example Transformation:**

| Before              | After           |
|---------------------|-----------------|
| `rec { x = 1; }`   | `{ x = 1; }`   |

## Changes

| File                                    | Description                                |
|-----------------------------------------|--------------------------------------------|
| `nixd/lib/Controller/CodeAction.cpp`    | Mark `DK_ExtraRecursive` as isPreferred    |

## Behavior

Action offered when:

- Cursor is on a `rec` keyword where no binding references another binding in the same attrset

Action NOT offered when:

- Bindings reference each other (e.g., `rec { x = 1; y = x; }`)

## Test Plan

- [x] All 3 extra-rec tests pass
- [x] Build with sanitizers succeeds
- Tests cover:
  - `basic.md`: Single-line `rec { x = 1; }` removal
  - `multiline.md`: Multiline attrset with independent bindings
  - `needed-rec.md`: **Negative test** — action not offered when `rec` is needed

## Related

- Issue: #466
- Draft PR: #755 (this implements Split 14)

### DIFF
--- a/nixd/lib/Controller/CodeAction.cpp
+++ b/nixd/lib/Controller/CodeAction.cpp
@@ -51,6 +51,7 @@ void Controller::onCodeAction(const lspserver::CodeActionParams &Params,
         bool IsPreferred = false;
         switch (D.kind()) {
         case nixf::Diagnostic::DK_UnusedDefLet:
+        case nixf::Diagnostic::DK_ExtraRecursive:
           IsPreferred = true;
           break;
         default:
--- a/nixd/tools/nixd/test/code-action/extra-rec/basic.md
+++ b/nixd/tools/nixd/test/code-action/extra-rec/basic.md
@@ -1,0 +1,78 @@
+# RUN: nixd --lit-test < %s | FileCheck %s
+
+Test basic `remove extra rec` action for unnecessary recursive attribute set.
+
+<-- initialize(0)
+
+```json
+{
+   "jsonrpc":"2.0",
+   "id":0,
+   "method":"initialize",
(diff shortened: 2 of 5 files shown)


nixd/lib/Controller/CodeAction.cpp
```diff
@@ -51,6 +51,7 @@ void Controller::onCodeAction(const lspserver::CodeActionParams &Params,
         bool IsPreferred = false;
         switch (D.kind()) {
         case nixf::Diagnostic::DK_UnusedDefLet:
+        case nixf::Diagnostic::DK_ExtraRecursive:
```
inclyc: There are also #786, #787, #788 PRs, they all just add `isPreferred = true` for some diagnostic ID, why those code actions from diagnostic `fix()` cannot be `preferred` by default?

takeokunn: Sorry, You're right — these PRs all add the same pattern and they conflict with #778.

#778 introduces `Fix::prefer()` so each fix declares itself as preferred at creation time, which removes this switch statement.

I'll close #785–#788, get #778 merged first, then open a single follow-up PR that adds `.prefer()` to the remaining diagnostic fixes (extra-rec, extra-with, empty-inherit, redundant-paren) together with their tests.

Issue #805, opened 2026-04-11
Missing Diagnostic For Invalid Enum NixOS Option Value


cw-1988:
Similar issue: [Missing Diagnostic For Invalid nullOr int NixOS Option Value](https://github.com/nix-community/nixd/issues/804)

"great first issue" label is missing, i would've assigned it. "low priority" could probably also be a good label for this kind of nitpick, but everything that forces us to reevaluate through nix eval is just an unnecessary waste of time.

`nixd` can resolve enum-valued NixOS options and show the allowed values for hover/completion, but does not report a diagnostic when a literal string outside the enum is assigned.

For example, `services.k3s.role` is typed as:

```text
one of "server", "agent"
```

Assigning another string does not produce an error diagnostic in the editor.

## Expected Behavior

Given this NixOS module:

```nix
{ ... }:

{
  services.k3s.role = "banana";
}
```

`nixd` should report a diagnostic for `"banana"`, because the option type is:

```text
one of "server", "agent"
```

Valid values would be:

```nix
services.k3s.role = "server";
services.k3s.role = "agent";
```

## Actual Behavior

No diagnostic is shown in the editor, even though `nixd` can resolve the option type information.

The invalid value is rejected when the NixOS module system evaluates the configuration:

```text
A definition for option `services.k3s.role' is not of type `one of "server", "agent"'.
Definition values:
- In `<unknown-file>': "banana"
```

## Reproduction

`flake.nix`:

```nix
{
  description = "nixd enum option value diagnostic reproduction";

  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-25.11";
  };

  outputs =
    { nixpkgs, ... }:
    {
      nixosConfigurations.control = nixpkgs.lib.nixosSystem {
        system = "x86_64-linux";
        modules = [
          ./configuration.nix
        ];
      };
    };
}
```

`configuration.nix`:

```nix
{ ... }:

{
  services.k3s.role = "banana";
}
```

VS Code / Cursor `nixd` config:

```json
{
  "nix.enableLanguageServer": true,
  "nix.serverPath": [
    "nixd",
    "--nixos-options-expr",
    "(builtins.getFlake \"path:/absolute/path/to/repro\").nixosConfigurations.control.options"
  ]
}
```

## Evaluation Verification

The NixOS module system rejects the value:

```bash
nix eval --impure --expr '
  let
    nixpkgs = builtins.getFlake "github:nixos/nixpkgs/nixos-25.11";
[...]

PR #809, opened 2026-04-12
feat(diagnostics): add wrong/empty enum Value diagnostic

cw-1988:
#805, completions on top are added in #810

AI slopped with GPT-5.4 High, but a basic DX feature i had to add.

Take your time, read at your own pace, this issue isnt going anywhere.

I will use my own fork with the features i deem necessary. I'm just here to share the goods. Make all the edits you want, or give me the necessary feedback so i can improve upon this PR.

I hope you are doing well and have a nice day ;-)


inclyc (commented):
This PR only validates whether enum values exist within the type system, but it skips other types in `lib.type` like `bool` or `string`. I'm concerned that checking only enums might be a bit 'heavy' in terms of performance impact. What do you think?

PR #836, opened 2026-06-07
Use `workspace/executeCommand` instead of `codeAction/resolve` to be spec compliant

### LINKED ISSUE #818: Unprompted opening of noogle.dev
### Describe the bug

At the beginning of the year, I got the following bug report for using `nixd` in IntelliJ at NixOS/nix-idea#98:

> Following https://github.com/nix-community/nixd/pull/764 when using a recent nixd (e.g. from nixpkgs/unstable) whenever the cursor is on a symbol from lib.* and ANY key or mouse button is pressed, a new browser tab opens with https://noogle.dev/f/lib/abc

(A workaround has been implemented at NixOS/nix-idea#100 by @Meallia.)

I looked at some documentation, and I think `nixd` is actually not following the LSP specification. Specifically, `nixd` seems to treat `codeAction/resolve` at [`Controller::onCodeActionResolve`](https://github.com/nix-community/nixd/blob/d37a39b1a53a5aa0e5e76cbe753b4228a0852a68/nixd/lib/Controller/CodeAction.cpp#L126) as a signal, that the code action has been executed. However, that does not match the documentation by Microsoft:

> The request is sent from the client to the server to resolve additional information for a given code action. This is usually used to compute the edit property of a code action to avoid its unnecessary computation during the `textDocument/codeAction` request.
> — [Language Server Protocol Specification - 3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction_resolve)

According to this documentation, `codeAction/resolve` only exists to allow for some lazy loading. It does not signal that the action is executed. While VSCode my only care about the `edit` property once it executes the actions, other IDEs my call this method early. Either because they do not implement lazy loading, or because they may for example want to show a preview of the changes.

### Suggestion

After scanning over the specification, I suspect a correct implementation would need to follow the following steps, but I might be wrong since I am not very familiar with the language server protocol.

- Define the availability of a command in the server capabilities:

  ```json
  {
    ...
    "executeCommandProvider": {
      "commands": ["noogle.dev"]
    },
    ...
  }
  ```

- Implement handler for the new command in [`workspace/executeCommand`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand) (This should match the current implementation of `codeAction/resolve`.)

- Provide the `"command"` in the `CodeAction` body (you no-longer need `"data"`):

  ```json
  {
    "title": "Open Noogle documentation for ...",
    "kind": "refactor",
    "command": {
      "title": "???",
      "command": "noogle.dev",
      "arguments": ["https://noogle.dev/..."]
    }
  }
  ```

### To Reproduce

1. Start IntelliJ
2. Install the plugin [nix-idea 0.4.0.19](https://github.com/NixOS/nix-idea/releases/tag/v0.4.0.19) or earlier
3. Use `nixd` as the language server in Settings → Languages & Frameworks → Nix → Language Server (LSP)
4. Navigate with the cursor to any usage of `lib.*` in a `*.nix` file

### Expected behavior

Allow for normal navigation using the mouse or arrow keys.

### Screenshots

_No response_

### Additional context

_No response_

JojOatXGME:
This PR replaces `codeAction/resolve` with `workspace/executeCommand` for action callbacks.

While this might not be the cleanest implementation, I should demonstrate how I think this functionality should be implemented to stay compliant to the LSP specification.

Fixes #818


JojOatXGME:
Looks like I would need to adjust some tests. Unfortunately, I am not sure how to run the tests locally because I run into compile errors unrelated to my changes caused by `-Werror=pedantic`. Not sure why this happens and how to fix it. (A build via `nix run .` works, though.)

PR #845, opened 2026-06-25
fix(code-action): preserve non-matching bindings in bulk Pack All

### LINKED ISSUE #837: "Pack all" removes unrelated attributes
```nix
{
  a.x = 1;
  b.y = 2;
  a.z = 3;
}
```
Running "Pack all 'a' bindings to nested set" yields:
```nix
{
  a = { x = 1; z = 3; };
}
```
Notice that `b.y = 2` was removed.

Version: /nix/store/31ns7kqnsw90p8knh44sjn9gqibjs4bi-nixd-2.9.1

inclyc:
The bulk "Pack all 'X' bindings to nested set" and "Recursively pack all ..." actions computed a single contiguous range from the leftmost to the rightmost matching sibling and replaced it with the packed text. When matching siblings were non-contiguous, that range engulfed intermediate non-matching bindings, which the replacement silently dropped.

For example, `{ a.x = 1; b.y = 2; a.z = 3; }` produced `{ a = { x = 1; z = 3; }; }` instead of preserving `b.y = 2;`.

Emit a multi-edit WorkspaceEdit instead: replace the first matching binding with the packed text, then delete each subsequent matching binding in place. Intermediate non-matching bindings are left untouched.

Tighten the existing non-contiguous and shallow-bulk lit tests to verify the two-edit shape, and add a regression test that mirrors the original bug report.

Fixes: https://github.com/nix-community/nixd/issues/837

### DIFF
--- a/nixd/lib/Controller/CodeActions/PackAttrs.cpp
+++ b/nixd/lib/Controller/CodeActions/PackAttrs.cpp
@@ -8,7 +8,7 @@
 
 #include <nixf/Basic/Nodes/Attrs.h>
 
-#include <optional>
+#include <vector>
 
 namespace nixd {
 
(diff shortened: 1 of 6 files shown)


MrQubo:
Could we make it so it doesn't leave empty lines?
E.g.
```nix
{
  a.x = 1;
  b = 2;
  a.y = {
    d = 3;
  };
  c = 4;
}

```
using code actions gives:
```nix
{
  a = { x = 1; y = {
    d = 3;
  }; };
  b = 2;
  
  c = 4;
}
```
but I think it would be better to produce:
```nix
{
  a = { x = 1; y = {
    d = 3;
  }; };
  b = 2;
  c = 4;
}

```

It shouldn't remove empty lines that were already present, so e.g.
```nix
{
  a.x = 1;
  a.y = 3;

}
```
should be transformed to:
```nix
{
  a = { x = 1; y = 3; };

}
```

Implementation wise I think, the deletion should extend to the beginning and to the end of the line (including end of the line), as long as it's only whitespace.Z

I'm asking for this, because nixfmt won't remove single empty lines, as it leaves empty lines on purpose.

MrQubo:
I don't want my above request to block this bugfix. I think this is good to be merged. I can make a separate issue for the request above.

Malix-Labs:
@MrQubo you can CC me in your new issue

Issue #317, opened 2024-01-24
Big closure size


ck3d:
The closure size of nixd is round about 550Mb on latest nixpkgs.

I would be great if the runtime dependency to llvm-dev and nix-dev can be removed.

inclyc:
> I would be great if the runtime dependency to llvm-dev and nix-dev can be removed.

AFAIK, we do not rely on llvm-dev & nix-dev as runtime dependency, only llvm.out & nix.out.

inclyc:
LLVM closure size is *much* larger than many other repositories (e.g. Debian), it's only about 30~40M if properly packaged[^llvm-debian]. You may refer to https://github.com/NixOS/nixpkgs/issues/230616 and these llvm static libraries contributed most closure size (which is definitely unnecessary at runtime).

[^llvm-debian]: For example, on Debian https://packages.debian.org/bookworm/llvm-14 it is only 48,353kb (47.2M).

ck3d:
You are right, the llvm itself is the main problem.

My mentioned packaged llvm-dev and nix-dev are only a problem on darwin, the dependencies are:

/nix/store/qlml6qifkgs47s3gb714lcq4vh27qpic-libcxxabi-16.0.6
/nix/store/6zn9dj3sdncl5qafw52p5zx7zszxjpjm-libcxx-16.0.6
/nix/store/a6kpihjkaiwxj1z544ha4s39mqgsjv8p-llvm-16.0.6-lib
/nix/store/4fkk4s69qln8nsbkd46n77s48cli1i6i-llvm-16.0.6-dev
/nix/store/909bg372n5iij7cs8vyggzkpb381dkla-boehm-gc-8.2.2
/nix/store/zic94z6xi9vrhd65z5dsn1q3bwab7wa8-nix-2.16.2
/nix/store/8mqyfmzb9nbfa3gghzfzkbv56zwzb44x-nix-2.16.2-dev
/nix/store/awwqpd7740g1m0qpxgcljmf18amsb1ky-libbacktrace-unstable-2022-12-16
/nix/store/8rsj7vmz86kgwy8aq8hbqg4kvsc22cw5-nixd-1.2.2

On linux, the package dependency shows following:

nix/store/p3jshbwxiwifm1py0yq544fmdyy98j8a-glibc-2.38-27
/nix/store/4zp05qkzh9s3c3k7pp9d36vhiz3x8h9l-gcc-13.2.0-lib
[...]

PR #854, opened 2026-07-11
build: add llvm_static option to statically link LLVM support

Allsimon:
nixd only uses LLVM's Support/ADT utilities (StringRef, JSON, Error, CommandLine, ...), yet linking against the monolithic libLLVM dylib pulls the entire LLVM lib output (~550 MiB) into the runtime closure.

Add an llvm_static meson option (default: false, no behavior change) that links just the static "support" component instead. Enable it in the Nix package, where the static archives are always available; this requires zlib and libxml2 at link time, which libLLVM previously resolved internally.

Shrinks the nixd runtime closure from 702 MiB to 158 MiB.

see https://github.com/nix-community/nixd/issues/317

### DIFF
--- a/nixd/default.nix
+++ b/nixd/default.nix
@@ -12,6 +12,7 @@
   boost,
   libxml2,
   zlib,
+  llvmStatic ? true,
 }:
 
 let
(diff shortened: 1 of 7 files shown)


inclyc (commented):
The direction itself LGTM; however, we should test this in CI to avoid silent failures.

Allsimon:
What kind of test do you want to add ? 

One checking that the runtime closure doesn't grow or building with both `llvm_static: true` and `llvm_static: false` ?

inclyc:
> What kind of test do you want to add ?
> 
> One checking that the runtime closure doesn't grow or building with both `llvm_static: true` and `llvm_static: false` ?

The latter. Because this flag is optional, I'd like to confirm that it's buildable in CI.

nixd/default.nix
```diff
@@ -10,6 +10,9 @@
   llvmPackages,
   gtest,
   boost,
+  libxml2,
+  zlib,
+  llvmStatic ? true,
```
inclyc: > Add an llvm_static meson option (default: false, no behavior change) 

Now there are indeed some users consuming this flake as their daily `nixd` configuration, so I would like the default `nixd` output here to remain consistent with the previous version. At the same time, this also aligns with the default value in Meson.

inclyc (approved):

inclyc (approved):
LGTM, thank you ❤️ @Allsimon

Issue #790, opened 2026-02-16
Autocompletion and Diagnostics for `self`


Ryokune:
**Is your feature request related to a problem? Please describe.**
Would be nice to have. Currently, I maintain a flake that uses flake parts and import tree, and I access imports via `self.nixosModules.foo` `self.nixosModules.myService` etc.
Ex:
```nix
... modules/hosts/myHost:
flake.nixosConfigurations.myHost = inputs.nixpkgs.lib.nixosSystem {
  modules = [
    self.nixosModules.someModule
  ]
}

... flake.nix:
outputs =
    inputs@{ self, ... }:
    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
      imports = [
        (inputs.import-tree ./modules)
      ];
      debug = true;
    };
```
Currently by how things are I could mistype something, or access an invalid attribute eg: `self.invalid.module` and nixd won't catch it via diagnostics. But i am unsure if this is just a misconfiguration on my end.

**Additional context**
Using hercules-ci:flake-parts and vic:import-tree in a project.

etrobert:
I run into this regularly when working on my flake-parts setup — `self.nixosModules.foo`, `self.packages.${system}.bar`, etc. are used throughout to wire things together, so not being able to jump to their definitions is a real friction point. Would love to see this land.

PR #856, opened 2026-07-14
libnixf/Sema: suppress flake-injected outputs formals

### LINKED ISSUE #649: allow flake boilerplate with unused `{outputs = { self }: { }`
Most typical flake boilerplates (e.g., [here](https://zero-to-nix.com/concepts/flakes/)) involve something like:

```nix
# flake.nix
{
  outputs = { self }: { }
}
``` 

This tiggers this nixd lint:

```
attribute `self` of argument is not used
```

This issue was previously raised in #589, where @inclyc suggested to just use ellipses `{ ... }` instead of self.

However:

- @inclyc warns that the semantics are not quite the same
- a lot of people with flake boilerplate will stub their toe on this

It'd be nice if either:
1. nixd shipped with some exception to now ring the alarm on this (unused `self`s).
2. if `...` is, in fact, safe/equivalent, ship that as an auto suggestion.

wasdhjklxyz:
Fixes #649

The Nix flake evaluator calls `outputs` as:

```nix
flake.outputs(inputs // { self = result; })
```

This means `self` and every name declared under `inputs` are unconditionally passed to `outputs` in `call-flake.nix`[^1].

[^1]: [call-flake.nix#L71](https://github.com/NixOS/nix/blob/40f375fac1af1a432cbf48dfd15468b1921d458c/src/libflake/call-flake.nix#L71)

This PR suppresses `sema-unused-def-lambda-noarg-formal` and `sema-unused-def-lambda-witharg-formal` for formals in the `outputs` lambda whose names are in `{ self } + keys(inputs)`, derived statically from the AST. Suppression is scoped to the `outputs` lambda's own formals in `flake.nix` specifically (nested lambdas and all other files are unaffected).

Broader flake support (go-to-definition on `self.*`, completion for `self.packages.*`) requires eval-layer work and is tracked separately in #790. Implementing this could make this fix redundant not sure though.


inclyc (changes requested):
These modifications to the Nix syntax/semantic layer should be made under the `libnixf/` directory. 

The only challenge might be that we need some way to determine whether the input file is named "flake.nix". I think hardcoding this option is probably feasible, since the community effectively already uses `flake.nix` and has reached a consensus on it.

wasdhjklxyz:
The only challenge might be that we need some way to determine whether the input file is named "flake.nix". I think hardcoding this option is probably feasible, since the community effectively already uses flake.nix and has reached a consensus on it.

libflake hardcodes flake.nix so I think we should do the same [1](https://github.com/nix-community/nixd/pull/856#user-content-fn-1-298623d256688356010303bd5389ce47)[2](https://github.com/nix-community/nixd/pull/856#user-content-fn-2-298623d256688356010303bd5389ce47)[3](https://github.com/nix-community/nixd/pull/856#user-content-fn-3-298623d256688356010303bd5389ce47). so instead of that suppression function in actOnDocumentAdd() maybe adding something like bool IsFlake = false as a parameter to runOnAst() which is called right before that? and later suppress the warning in emitEnvLivenessWarning() if the formal is in that set and scoped to the outputs lambda?

Footnotes
[flake.hh](https://github.com/NixOS/nix/blob/40f375fac1af1a432cbf48dfd15468b1921d458c/src/libflake/include/nix/flake/flake.hh#L87) [↩](https://github.com/nix-community/nixd/pull/856#user-content-fnref-1-298623d256688356010303bd5389ce47)

[call-flake.nix](https://github.com/NixOS/nix/blob/40f375fac1af1a432cbf48dfd15468b1921d458c/src/libflake/call-flake.nix#L65) [↩](https://github.com/nix-community/nixd/pull/856#user-content-fnref-2-298623d256688356010303bd5389ce47)

[flake-ref.cc](https://github.com/NixOS/nix/blob/40f375fac1af1a432cbf48dfd15468b1921d458c/src/libflake/flakeref.cc#L129) [↩](https://github.com/nix-community/nixd/pull/856#user-content-fnref-3-298623d256688356010303bd5389ce47)

inclyc:
>> The only challenge might be that we need some way to determine whether the input file is named "flake.nix". I think hardcoding this option is probably feasible, since the community effectively already uses flake.nix and has reached a consensus on it.
>
> libflake hardcodes flake.nix so I think we should do the same.

OK.


> so instead of that suppression function in actOnDocumentAdd() maybe adding something like bool IsFlake = false as a parameter to runOnAst() which is called right before that? 

Yes.

> and later suppress the warning in emitEnvLivenessWarning() if the formal is in that set and scoped to the outputs lambda?

Yes, exactly.

inclyc (approved):
LGTM, thank you!

PR #860, opened 2026-07-22
fix(completion): retrigger incomplete results

mjrogozinski:
## Summary

- mark empty completion responses triggered by `.` as incomplete
- mark evaluator-capped option responses as incomplete
- preserve the controller's `CheckDefault` and early-return patterns
- add/update LSP regressions for both incomplete-result paths

## Root cause

nixd advertises `.` as a completion trigger. For an incomplete selector such as `pkgs.`, parser recovery can reject the completion position and return an empty `CompletionList` with `isIncomplete: false`.

Option completion had a related pagination problem. The evaluator stopped after exactly 30 matching options, while the controller only recognized truncation when an additional item exceeded its matching 30-item limit. A capped response was therefore reported with `isIncomplete: false` even though narrower prefixes could produce candidates outside the first page.

Clients such as blink.cmp cache complete responses. They consequently kept filtering the empty or truncated result as the user typed instead of sending a prefix-specific request.

## Fix

After the controller has produced a response, an empty dot-triggered result is marked incomplete. This covers parser-recovery early exits without replacing the existing `CheckDefault` idiom or changing the switch's early-return structure.

The option evaluator now returns one item beyond its limit as a truncation sentinel, matching package completion. The controller drops that extra item at its own limit and sets `isIncomplete: true`, prompting `TriggerForIncompleteCompletions` requests as the prefix grows.

Manual requests, non-empty dot results, and option sets that fit within the limit retain their existing completeness behavior.

## Validation

- Built with `nix develop -c meson compile -C build`
- Full Meson suite passes (6/6 targets, including the nixd LSP regressions)
- `clang-format --dry-run --Werror` passes
- `git diff --check` passes
- Raw Neovim LSP probe against the packaged binary confirms a capped 30-item option response is now incomplete and a subsequent prefix retrigger reaches nixd
- Verified interactively with Neovim and blink.cmp that `pkgs.pyt` includes `python312`

### DIFF
--- a/nixd/lib/Controller/Completion.cpp
+++ b/nixd/lib/Controller/Completion.cpp
@@ -465,9 +465,10 @@ void Controller::onCompletion(const CompletionParams &Params,
                               Callback<CompletionList> Reply) {
   using CheckTy = CompletionList;
   auto Action = [Reply = std::move(Reply), URI = Params.textDocument.uri,
-                 Pos = toNixfPosition(Params.position), this]() mutable {
+                 Pos = toNixfPosition(Params.position),
+                 Context = Params.context, this]() mutable {
     const auto File = URI.file().str();
-    return Reply([&]() -> llvm::Expected<CompletionList> {
+    auto Result = [&]() -> llvm::Expected<CompletionList> {
       const auto TU = CheckDefault(getTU(File));
       const auto AST = CheckDefault(getAST(*TU));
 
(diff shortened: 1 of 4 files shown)


inclyc:
I have refactored some fixes in #888 + #887

mjrogozinski:
> I have refactored some fixes in #888 + #887

Thank you! I will try to run the latest version on my machine

PR #878, opened 2026-08-23
vim inlay hint

evanwporter:
I wanted to display the versions of my vim plugins like so:

image

Note: This PR is entirely AI generated but I made this for myself mostly so I’m okay if you reject this

One possible issue is that `mkSelector` can throw, and I don't catch it. But is there something that may catch it up the stack? 

Ideally rather than throwing it should be wrapped in something like std::optional or std::expected, but I'm still new to the repo and I don't really what's one docket for the future.

But yeah generally doing this would lower the amount of crashes I think.

Also I just **_really_** don't like using exceptions.

### DIFF
--- a/nixd/lib/Controller/InlayHints.cpp
+++ b/nixd/lib/Controller/InlayHints.cpp
@@ -56,6 +56,32 @@ class NixpkgsInlayHintsProvider {
 
   llvm::StringRef Src;
 
+  void addHint(const Node &N, Selector Selector) {
+    if (!rangeOK(N.positionRange()))
+      return;
+
+    // Ask nixpkgs eval to provide its information. This is relatively slow.
+    // Maybe better query a set of packages in the future?
+    std::binary_semaphore Ready(0);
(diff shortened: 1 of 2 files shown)


inclyc (commented):
The direction seems to be good, please wait for details review.

evanwporter:
@inclyc I’m closing this because I made it in less than ten minutes with ai and I don’t think it’s worth wasting your time since you could’ve done the same.

I’m currently using my own build of nixd with this patch included though. 

Crazy times we live in lol.

But I really like nixd, and honestly i think it’s super impressive what you’ve done here (ie: parsing the entire semantic tree of the nix language). 

My biggest gripe with this language server, is that it crashes quite a bit (also that it uses meson which is a foreign language to me lol especially after I struggled to learn cmake) The thing is I’m having trouble reproduce the crashes. 

But anyways I’d love to get involved with this project more in the future when hopefully I have more free time to send well crafted prs.

